PR #21, opened 2021-05-18
update release.yaml to include docs build and release

alonashkenazi:


.github/workflows/release.yaml
```diff
@@ -56,4 +56,25 @@ jobs:
...
+          python-version: '3.9'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install mkdocs mkdocs-material mkdocstrings
+
+      - name: Deploy docs
+        run: |
+          mkdocs gh-deploy
```
yi2020, line 80: This uses the `gh-pages` branch. Maybe we should have it "protected"? While we're at it, we should protect `main` as well?

alonashkenazi: I will do it separately (no PR is needed)

yi2020 (approved):

PR #29, opened 2021-05-19
Cr 2150 - Extract excessive logic from rules

Tomerva:


cloudrail/knowledge/context/mergeable.py
```diff
@@ -84,3 +84,7 @@ def is_managed_by_iac(self) -> bool:
     @abstractmethod
     def is_tagable(self) -> bool:
         pass
+
+    # pylint: disable=no-self-use
```
Imanuel-Mizrahi, line 88: @Tomerva , did you meant to add this to pylintrc file?

Imanuel-Mizrahi (approved):

alonashkenazi (approved):

cloudrail/knowledge/context/mergeable.py
Tomerva: no, only for this function (its a wrong pylint error indicating im not using 'self' so it should be class/static method, but its abstract so ofcourse its not used here)

PR #46, opened 2021-05-26
39 more doc updates

yi2020:


docs/context/aws/apigateway/README.md
```diff
@@ -1,3 +1,10 @@
+## Sample rules
```
alonashkenazi, line 1: for all the samples, can you use snippet to include the code in the doc page so the reader will not need to redirect from this page?
https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#snippets

alonashkenazi (changes requested):

docs/context/aws/apigateway/README.md
yi2020: Good idea, done:

Screen Shot 2021-05-26 at 12 41 43 PM
Screen Shot 2021-05-26 at 12 41 36 PM

alonashkenazi (approved):

PR #47, opened 2021-05-26
Changed EnsureSecurityGroupIncludeDescriptionRule.should_run_rule to …

Tomerva:
…return true if there are security groups. improved security_group.used_by by checking that ENI's that use this SG are valid and are also in use

### DIFF
--- a/cloudrail/knowledge/context/aws/ec2/network_interface.py
+++ b/cloudrail/knowledge/context/aws/ec2/network_interface.py
@@ -2,7 +2,6 @@ from typing import List, Optional
 
 from cloudrail.knowledge.context.aws.aws_connection import ConnectionInstance
 from cloudrail.knowledge.context.aws.aws_resource import AwsResource
-from cloudrail.knowledge.context.aws.ec2.security_group import SecurityGroup
 from cloudrail.knowledge.context.aws.service_name import AwsServiceName
 
 
@@ -54,7 +53,7 @@ class NetworkInterface(ConnectionInstance, AwsResource):
         self.is_primary: bool = is_primary
         self.availability_zone: str = availability_zone
         self.subnet: 'Subnet' = None
-        self.security_groups: List[SecurityGroup] = []
+        self.security_groups: List['SecurityGroup'] = []
         self.owner: Optional[AwsResource] = None
         self.aliases.add(eni_id)
 
@@ -110,6 +109,6 @@ class NetworkInterface(ConnectionInstance, AwsResource):
     def is_tagable(self) -> bool:
         return True
 
-    def add_security_group(self, security_group: SecurityGroup):
+    def add_security_group(self, security_group: 'SecurityGroup'):
         self.security_groups.append(security_group)
         security_group.add_usage(self)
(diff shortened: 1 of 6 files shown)


Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/aws/ec2/security_group.py
```diff
@@ -37,7 +38,8 @@ def __init__(self, security_group_id: str, region: str, account: str,
 
     @property
     def used_by(self) -> Set[AwsResource]:
-        return {resource for resource in self._used_by if not resource.invalidation}
+        return {resource for resource in self._used_by if not resource.is_invalidated and
```
alonashkenazi, line 41: please add unit test. thx

Tomerva, line 41: @yi2020 i have a hard time deciding what to do here..
security_group.used_by is a field that holds the resources (so far ENI's) that use this security group. this is used for rules that check if a security group was used.

Should this set contain only valid resources or can it contain invalid resources?

Right now im leaning toward allowing invalid resources to be part of the used_by property. This is because when a security group is attached to an ENI, and that ENI is attached to a resource, then it means that the SG is used regardless on if we have all the data of the ENI/resource (which could invalidate them).

WDYT?

So to recap, its either not allowing invalidated resource/ENI to be part of the used_by field (the above implementation)
or allow invalid resource (and the implementation will be `if not isinstance(resource, NetworkInterface) or resource.owner`

Tomerva: "Which in turn, will mean the security group is pointing at something invalid" - no, used_by is excluded from the invalidation process. We dont want to invalidate a security group because its attached to an invalid resource, because it could also be attached to some other resource..

yi2020:
Ah good point. And so, you may run into a situation where a security group
will only point to invalid resources. If you remove those resources from
used_by, then used_by may be empty, even though it shouldn't be.

However, we also guarantee a consistent context. All content in the context
should point to valid resources, right? So it makes sense to remove invalid
resources from used_by. If a security group is used by resources, but all
of them are invalid, we should remove the security group.

cloudrail/knowledge/context/aws/ec2/security_group.py
Tomerva: Ok i agree, except for the part where we should remove the security group, because it doesnt actually mean its not valid

alonashkenazi (approved):

cloudrail/knowledge/context/aws/ec2/network_interface.py
```diff
@@ -54,7 +53,7 @@ def __init__(self,
         self.is_primary: bool = is_primary
         self.availability_zone: str = availability_zone
         self.subnet: 'Subnet' = None
-        self.security_groups: List[SecurityGroup] = []
+        self.security_groups: List['SecurityGroup'] = []
```
yi2020, line 56: @Tomerva huh?

PR #48, opened 2021-05-26
CR-2190: [Not CAR] [AWS] Ensure RDS clusters have a backup retention policy

Imanuel-Mizrahi:
added context rule and tests


Imanuel-Mizrahi:
No need in this PR, already merged as part of another one.

PR #51, opened 2021-05-30
CR-2098-fix_should_run_rule_to_remove_account

Imanuel-Mizrahi:
modified needed rules, to not use account as mandate


Imanuel-Mizrahi:
Following this code change, all unit-tests passed with ratio = 0:https://jenkins.aws.indeni-ops.com/job/Cloudrail-unit-tests/506/

cloudrail/version.py
```diff
@@ -1 +1 @@
-__version__ = 'v0.0.24'
+__version__ = 'v0.0.25-beta.1'
```
alonashkenazi, line 1: you shouldn't commit beta

alonashkenazi (changes requested):

cloudrail/version.py
Imanuel-Mizrahi: @alonashkenazi , this was done automatically.
How can I avoid that?

alonashkenazi: commit __version__ = 'v0.0.24' back

alonashkenazi (approved):

Imanuel-Mizrahi:
Tested this changes using cloudrail CLI, all affected rules listed in the ticket - passed.

Tomerva (approved):

cloudrail/knowledge/context/aws/iam/iam_password_policy.py
```diff
@@ -45,7 +45,10 @@ def get_keys(self) -> List[str]:
         return [self.account]
 
     def get_name(self) -> str:
-        return self.account + ' policy'
+        if self.account:
+            return self.account + ' policy'
+        else:
+            return 'The account'
```
yi2020, line 51: Should this be "the account's policy"?

yi2020 (changes requested):

cloudrail/knowledge/context/aws/iam/iam_password_policy.py
Imanuel-Mizrahi: Worth thinking of it. we are dealing with the IAM account pass policy.
Perhaps we should change it to:
if self.account:
  return f'IAM account password policy for account {self.account}'
else:
  return 'IAM account password policy'
?

yi2020: That works

Imanuel-Mizrahi: OK @yi2020 , change made

yi2020 (approved):

PR #55, opened 2021-06-02
FunctionAppAuthenticationEnableRule: new azure rule

idan-aviv:


cloudrail/version.py
```diff
@@ -1 +1 @@
-__version__ = 'v0.0.26'
+__version__ = 'v0.0.27-beta.4'
```
Imanuel-Mizrahi, line 1: as far as I know, we do not push beta, so that should be changed

Imanuel-Mizrahi (approved):

cloudrail/version.py
idan-aviv: @Imanuel-Mizrahi i didn't change it, its automatically added by github flow when you deploy to pipy

cloudrail/knowledge/context/azure/azure_resources/web_app/azure_function_app.py
```diff
@@ -0,0 +1,35 @@
+from typing import Optional, List
+from cloudrail.knowledge.context.azure.azure_resources.azure_resource import AzureResource
+from cloudrail.knowledge.context.azure.azure_resources.constants.azure_resource_type import AzureResourceType
+from cloudrail.knowledge.context.azure.azure_resources.web_app.auth_settings import AuthSettings
+
+
+class AzureFunctionApp(AzureResource):
```
yi2020, line 7: Missing docs

cloudrail/knowledge/rules/azure/non_context_aware/function_app_authentication_enable_rule.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class FunctionAppAuthenticationEnableRule(AzureBaseRule):
+
+    def get_id(self) -> str:
+        return 'non_car_authentication_enabled_in_function_app'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for func_app in env_context.function_apps.values():
```
yi2020, line 16: Why do you need .values() ? When using the context in AWS, we don't need .values()

yi2020: (or in other words, why use AliasesDict)

cloudrail/knowledge/context/azure/azure_resources/web_app/azure_function_app.py
```diff
@@ -0,0 +1,35 @@
+from typing import Optional, List
```
yi2020, line 1: Can you `git mv` everything under azure_resources to simply azure, so it matches the directory structure we have for `/context/aws`? I know it's not part of this rule, but I've already asked for this in another PR and it wasn't done.

yi2020 (changes requested):

cloudrail/knowledge/rules/azure/non_context_aware/function_app_authentication_enable_rule.py
Tomerva: keep using AliasesDict, theres just no need to do .values() when iterating

cloudrail/version.py
Tomerva: when you do release, it automatically modifies version.py to include the new version

yi2020 (approved):

PR #56, opened 2021-06-02
Cr 2219

Tomerva:


cloudrail/knowledge/context/aws/ec2/security_group.py
```diff
@@ -37,9 +37,13 @@ def __init__(self, security_group_id: str, region: str, account: str,
...
-    def used_by(self) -> Set[AwsResource]:
-        return {resource for resource in self._used_by if not resource.is_invalidated and
-                (not isinstance(resource, NetworkInterface) or (resource.owner and not resource.owner.is_invalidated))}
+    def is_used(self) -> bool:
+        """
+        This property returns true if any resource is using this security group, including invalidated resources.
+
+        Note: Network interfaces that use this security group, but are not attached to a resource, will not count.
+        """
+        return any(resource for resource in self._used_by if not isinstance(resource, NetworkInterface) or resource.owner)
```
yi2020, line 46: I thought we discussed using a flag, so that if invalidation removes the interfaces from this resource is_used will still be True?

yi2020 (changes requested):

cloudrail/knowledge/context/aws/ec2/security_group.py
Tomerva: yeah, if the attached ENI or its owner are invalid the result will still be True

Tomerva: just to be clear, invalidation process doesnt remove the link between resources, it removes the invalidated resource from the context object

so for example if you have a Subnet and it has a link to a Vpc thats invalidated, subnet will be removed from environment_context.subnets, vpc will be removed from environment_context.vpcs, but subnet.vpc will remain

yi2020:
Oh I didn't realize that. It wasn't a problem because both vpc and subnet
are removed, so it's not possible to get from the vpc to the subnet or vice
versa. Understood.

yi2020 (approved):

alonashkenazi (approved):

PR #63, opened 2021-06-07
Cr 2256 2

Tomerva:


cloudrail/knowledge/context/azure/webapp/azure_app_service.py
```diff
@@ -27,9 +26,6 @@ def get_cloud_resource_url(self) -> Optional[str]:
         return f'https://portal.azure.com/#@{self.tenant_id}/resource/subscriptions/{self.subscription_id}/resourceGroups/' \
                f'{self.resource_group_name}/providers/Microsoft.Web/sites/{self.name}/appServices'
 
-    def get_friendly_name(self) -> str:
-        return self.get_name()
```
Viperain, line 31: what self.get_name() ? why do we need it? cant we use self.name?

Viperain (approved):

PR #64, opened 2021-06-07
Rename Rules category in docs

yi2020:
Closes #57

PR #66, opened 2021-06-08
Add azure context docs and missing AWS nav items

yi2020:


yi2020:
@alonashkenazi same livereload error as in #64 . It's odd as this works locally with the same requirements and I don't think anything has changed.

alonashkenazi (approved):

PR #68, opened 2021-06-08
Cr 2142 vpn rule

Tomerva:


cloudrail/knowledge/context/azure/network/azure_vnet_gateway.py
```diff
@@ -0,0 +1,39 @@
+from enum import Enum
```
yi2020, line 1: Please add this file to the Azure context documentation, per message in Slack on #cloudrail-knowledge-developers

yi2020 (changes requested):

cloudrail/knowledge/context/azure/network/azure_vnet_gateway.py
Tomerva: Fixed, please re-review

Viperain (approved):

PR #76, opened 2021-06-09
CR-1264: [Non-CAR][AWS] Lambda function can be invoked by anonymous or all AWS authenticated users

Imanuel-Mizrahi:
rule ready


yi2020 (changes requested):

cloudrail/knowledge/rules/aws/non_context_aware/ensure_lambda_function_cannot_be_invoked_public_rule.py
```diff
@@ -0,0 +1,36 @@
...
+            return issues
+
+    def should_run_rule(self, environment_context: AwsEnvironmentContext) -> bool:
+        return bool(environment_context.lambda_function_list)
+
+    @staticmethod
+    def _is_lambda_can_publicly_invoked(lambda_policy: Policy) -> bool:
+        return any(statement.effect == StatementEffect.ALLOW
+                   and not statement.condition_block
+                   and any(action in ('*', 'lambda:InvokeFunction', 'lambda:*') for action in statement.actions)
```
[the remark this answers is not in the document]
Imanuel-Mizrahi: @yi2020 , I have changed it, to use get_intersected_actions() function, please review.
Thanks!

yi2020 (approved):

cloudrail/knowledge/context/aws/lambda_/lambda_policy_statements.py
```diff
@@ -42,3 +39,10 @@ def get_arn(self) -> str:
     @property
     def is_tagable(self) -> bool:
         return False
+
```
Imanuel-Mizrahi, line 42: @idan-aviv , please take a look at the changes I have made here.
and also in the relations assigner here:
https://bitbucket.org/indeni/cloud-rail/pull-requests/1415#chg-src/core/aws/aws_relations_assigner.py
Thanks!

cloudrail/knowledge/utils/arn_utils.py
```diff
@@ -105,3 +105,41 @@ def is_valid_arn(arn: str) -> bool:
     except Exception:
         logging.warning(f'failed parse arn {arn}')
         return False
+
+
+@functools.lru_cache(maxsize=None)
+def are_arns_intersected_for_tf(resource_arn: str, target_arn: str):
```
Tomerva, line 111: this code is duplicated
remove are_arns_intersected, and rename this function to are_arns_intersected

then all thats left is:
resource_arn_parsed = arnparse(resource_arn) if is_valid_arn(resource_arn) else DummyArnObject(resource_arn)
and same for target_arn_parsed

Imanuel-Mizrahi: Actually, it's almost true.
They are not 100% identical, there is also another line which I do not run for TF:
        elif not (is_valid_arn(resource_arn) and is_valid_arn(target_arn)):
            return False
I was thinking about adding a flag to the function, that will run this if condition only if it's TF.
What do you think @Tomerva ?

Imanuel-Mizrahi: @Tomerva , looking at the usage of "are_arns_intersected" in a few places inside the project, it might be better to add the resource object itself, in order to check if it's TF or not, and then decide.
Might need to zoom on that and decide...

cloudrail/knowledge/context/aws/lambda_/lambda_function.py
```diff
@@ -59,7 +59,7 @@ def get_name(self) -> str:
     def get_arn(self) -> str:
         return self.arn
 
-    def is_arn_match(self, arn: str):
+    def is_arn_match_for_lambda(self, arn: str):
```
Tomerva, line 62: there is no need to create a new function called is_arn_match_for_lambda
instead, keep it is_arn_match. then when you have a resource and you want to do resource.is_arn_match, you wont have to 'know' if resource is of type lambda

LMK if you need explanation on how this works

cloudrail/knowledge/rules/aws/non_context_aware/ensure_lambda_function_cannot_be_invoked_public_rule.py
```diff
@@ -0,0 +1,37 @@
...
+    def execute(self, env_context: AwsEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+
+        for lambda_func in env_context.lambda_function_list:
+            if self._is_lambda_can_publicly_invoked(lambda_func.resource_based_policy):
+                issues.append(
+                    Issue(
+                        f'The {lambda_func.get_type()} `{lambda_func.get_friendly_name()}` is exposed because its resource policy is too permissive.',
+                        lambda_func, lambda_func))
+            return issues
```
Tomerva, line 25: indentation is not good

cloudrail/knowledge/context/aws/lambda_/lambda_function.py
Imanuel-Mizrahi: Won't it be an issue, if the resource is defined as "AwsResource", and then it will take the "is_arn_match" function for "AwsResource" class, instead of Lambda_function class?

Imanuel-Mizrahi: @Tomerva , please see my query here. Thanks!

Tomerva (approved):

idan-aviv (approved):

cloudrail/knowledge/rules/aws/non_context_aware/ensure_lambda_function_cannot_be_invoked_public_rule.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def should_run_rule(self, environment_context: AwsEnvironmentContext) -> bool:
+        return bool(environment_context.lambda_function_list)
+
+    @staticmethod
+    def _is_lambda_can_publicly_invoked(lambda_policy: Policy) -> bool:
+        return any(statement.effect == StatementEffect.ALLOW
+                   and not statement.condition_block
+                   and get_intersected_actions(statement.actions, 'lambda:InvokeFunction')
```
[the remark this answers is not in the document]
Imanuel-Mizrahi: @idan-aviv , please check the ticket https://indeni.atlassian.net/browse/CR-1264, the conditions were set according to the logic inside the ticket.
This is not only "publicly" invoked, we also would like to avoid all AWS authenticated users and services, the name of this function may be misleading, as well as the name of the rule...
For the moment, I will not change it, as the conditions are fixed per the logic in the ticket, and should be used specifically here, so I will not move it as a property inside the lambda_function object.

cloudrail/knowledge/rules/aws/non_context_aware/ensure_lambda_function_cannot_be_invoked_public_rule.py
```diff
@@ -0,0 +1,37 @@
...
+                issues.append(
+                    Issue(
+                        f'The {lambda_func.get_type()} `{lambda_func.get_friendly_name()}` is exposed because its resource policy is too permissive.',
+                        lambda_func, lambda_func))
+            return issues
+
+    def should_run_rule(self, environment_context: AwsEnvironmentContext) -> bool:
+        return bool(environment_context.lambda_function_list)
+
+    @staticmethod
```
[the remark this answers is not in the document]
Imanuel-Mizrahi: replied inside the next comment

PR #102, opened 2021-06-27
Feature/cr 2229 align cloud and tf context

Imanuel-Mizrahi:


cloudrail/knowledge/context/aws/ec2/vpc.py
```diff
@@ -54,18 +54,19 @@ class Vpc(AwsResource):
...
                  friendly_name: str = None,
                  is_default: bool = False,
-                 tf_resource_type: AwsServiceName = AwsServiceName.AWS_VPC,
                  enable_dns_support: bool = None,
                  enable_dns_hostnames: bool = None):
-        super().__init__(account, region, tf_resource_type)
+        super().__init__(account, region, AwsServiceName.AWS_VPC)
         self.vpc_id: str = vpc_id
         self.aliases.add(vpc_id)
         self.cidr_block: List[str] = cidr_block
```
yi2020, line 68: I'm not familiar enough with Python but wouldn't the change to Optional risk breaking rules? For example, if a rule isn't aware this field may be None? This comment is true for all places where you added Optional on existing fields.

cloudrail/knowledge/utils/port_utils.py
```diff
@@ -413,3 +414,17 @@ def get_overlapping_ports(first_range: Tuple[int, int], second_range: Tuple[int,
         return None
     else:
         return max_start, min_end
+
+
+def convert_ip_protocol_to_protocol_name(ip_protocol_num: Union[str, int]) -> Optional[str]:
+    if ip_protocol_num in ('-1', -1):
+        return 'all'
```
yi2020, line 421: Is this "all" string an AWS string? What do Azure and GCP use? (note this function is not in an AWS-specific package)

cloudrail/knowledge/context/aws/ec2/vpc.py
Imanuel-Mizrahi: If a variable is None, and we are not aware of it, the rule might crash, but this is unrelated to the "optional".
We need to build the rule in a way that will consider the fact a variable might be None in some cases, unless this is not possible (as it's a required field, or something like this).

cloudrail/knowledge/utils/port_utils.py
Imanuel-Mizrahi: 'all' string is used in AWS console, not in CLI, or the API outputs.
I just thought it's more readable than '-1'.
GCP - use empty field for protocol to specify all ports.
Azure - use "Any" to specify all ports.

Should I change the name of the function to something with AWS, so it will be clear this is only for AWS relevant?

cloudrail/knowledge/context/aws/apigateway/api_gateway_stage.py
```diff
@@ -36,9 +38,13 @@ def __init__(self,
         self.stage_name: str = stage_name
         self.xray_tracing_enabled: bool = xray_tracing_enabled
         self.access_logs: Optional[AccessLogsSettings] = access_logs
+        self.method_settings: Optional[ApiGatewayMethodSettings] = None
 
     def get_keys(self) -> List[str]:
-        return [self.api_gw_id]
+        if self.method_settings:
+            return [self.api_gw_id, 'with_method_settings']
```
Tomerva, line 45: why is this necessary? api_gw_id is already unique

Imanuel-Mizrahi: I wanted to distinguish between stage which include method settings, and one which does not.

cloudrail/knowledge/context/aws/ec2/subnet.py
```diff
@@ -30,9 +30,8 @@ def __init__(self,
                  map_public_ip_on_launch: bool,
                  region: str,
                  account: str,
-                 is_default: bool = False,
-                 tf_resource_type: AwsServiceName = AwsServiceName.AWS_SUBNET):
-        super().__init__(account, region, tf_resource_type)
+                 is_default: bool = False):
+        super().__init__(account, region, AwsServiceName.AWS_SUBNET)
```
Tomerva, line 34: tf_resource_type was showing as a drift?

cloudrail/knowledge/context/aws/ec2/vpc.py
Tomerva: Yoni, just FYI, python is very dynamic, in fact its so dynamic that it doesnt really care what type you give it.. Optional just tells the developers that this might be none.

With that being said, what is the benefit of making this specific field optional? i think its much safer and easier to use as a non-optional list. meaning, if you dont have ipv6, just make this an empty list.. that way if we iterate over it, we wont need to check if its none or not.
semantically, empty list and None have basically the same, but empty list is safer.

cloudrail/knowledge/context/aws/ec2/subnet.py
Imanuel-Mizrahi: I think it did, not sure for which resource, as it was part of the builder I think it did.

cloudrail/knowledge/context/aws/elb/load_balancer_attributes.py
```diff
@@ -30,11 +30,11 @@ def __init__(self,
                  account: str,
                  region: str,
                  load_balancer_arn: str,
-                 drop_invalid_header_fields: bool,
+                 drop_invalid_header_fields: Optional[bool],
```
Tomerva, line 33: why is it optional? mostly boolean fields are resolveable at build time (theres always a default value..)

Imanuel-Mizrahi: You are right, I should change the builder, to return False as default.

cloudrail/knowledge/context/aws/ec2/subnet.py
Tomerva: @idan-aviv could this be an issue? this means that tf_resource_type for a subnett is always AWS_SUBNET and never AWS_DEFAULT_SUBNET

cloudrail/knowledge/utils/port_utils.py
```diff
@@ -413,3 +414,17 @@ def get_overlapping_ports(first_range: Tuple[int, int], second_range: Tuple[int,
...
+    elif not ip_protocol_num.isdigit():
+        return ip_protocol_num
+    else:
+        ip_protocol_table = {num: name.replace('IPPROTO_', '') for name, num in vars(socket).items() if name.startswith("IPPROTO")}
+        try:
+            return ip_protocol_table[int(ip_protocol_num)].lower()
+        except Exception as ex:
+            print(f'fail converting ip_protocol number for {ip_protocol_num} : {ex}')
```
Tomerva, line 429: use logger instead of print

cloudrail/knowledge/utils/port_utils.py
Tomerva: The only difference should be with the 'all' identifier. basically, the protocol can be represented by a number, and there is no unified representation of 'all", so each provider did whatever they like.

this function is a good solution for https://indeni.atlassian.net/browse/CR-2385
just need to define 'all' properly for all scenarios like: `if ip_protocol_num in (-1, '-1', 'Any', None, '')` or something like that

cloudrail/knowledge/utils/port_utils.py
Viperain: agree , u should never do print()
i dont understand what u do incase of exception. for now, it seems your return none. u need to either raise the exception and crash the process or return value the system can work with (For example all or 0 or -1)

Imanuel-Mizrahi: @Tomerva @Viperain , I have changed it to the following:
        except Exception as ex:
            logging.exception(f'fail converting ip_protocol number for {ip_protocol_num} : {ex}')
        return str(ip_protocol_num)

cloudrail/knowledge/context/aws/elb/load_balancer_attributes.py
Imanuel-Mizrahi: @Tomerva , I have removed the Optional, consider this conversation as resolved

yi2020 (approved):

yi2020 (approved):

Viperain (approved):

PR #104, opened 2021-06-29
Cr 2515 gcp

Tomerva:


cloudrail/knowledge/context/gcp/gcp_resource.py
```diff
@@ -0,0 +1,34 @@
...
+
+from cloudrail.knowledge.context.gcp.constants.gcp_resource_type import GcpResourceType
+from cloudrail.knowledge.context.mergeable import Mergeable
+
+
+class GcpResource(Mergeable):
+
+    def __init__(self, resource_type: GcpResourceType):
+        super().__init__()
+        self.tf_resource_type: GcpResourceType = resource_type
```
alonashkenazi, line 12: I think that tf_ is misleading here, Tomer how do you see GCP support other IaC languages?

cloudrail/knowledge/context/gcp/sql/gcp_sql_database_instance.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @property
+    def is_tagable(self) -> bool:
+        return False
+
+    def get_id(self) -> str:
+        return self.name
+
+    def get_cloud_resource_url(self) -> Optional[str]:
+        return f'https://console.cloud.google.com/sql/instances/{self.name}/'
```
alonashkenazi, line 30: should we have this in a const? https://console.cloud.google.com/sql/instances

cloudrail/knowledge/rules/rule_metadata.py
```diff
@@ -85,7 +85,13 @@ def rule_matches_query(rule_id: str, rule_name: str, query: Optional[str]) -> bo
 
 
 def get_rule_metadata_content(provider: CloudProvider) -> dict:
-    provider_name = 'aws' if provider == CloudProvider.AMAZON_WEB_SERVICES else 'azure'
+    if provider == CloudProvider.AMAZON_WEB_SERVICES:
```
alonashkenazi, line 88: I would prefer to throw an exception if the provider is missing and don't default to GCP

alonashkenazi (changes requested):

cloudrail/knowledge/context/gcp/gcp_resource.py
Tomerva: I took another look on what tf_resource_type stands for (since its originated from AwsResource)

as i see it, we should build the iac_state differently to include: IAC provider (tf, CFN, etc) and the block type (aws_vpc, azurerm_vnet, etc..). Also, i dont think its correct for the AwsResource/AzureResource/GcpResource to have the resource type as a parameter for init, since we can do it generically.

We can discuss this further tomorrow. But either way, i think we need a new ticket for this change because its not a gcp-related change

cloudrail/knowledge/context/gcp/sql/gcp_sql_database_instance.py
Tomerva: i added GcpResource._BASE_URL = https://console.cloud.google.com

cloudrail/knowledge/context/gcp/gcp_resource.py
alonashkenazi: Can you open a ticket for that?

cloudrail/knowledge/rules/rule_metadata.py
alonashkenazi: @Tomerva what do you think about this?

alonashkenazi (approved):

cloudrail/knowledge/rules/rule_metadata.py
Tomerva: i can add to CloudProvider this method:

`    def to_shorthand_string(self):
        if self == CloudProvider.AMAZON_WEB_SERVICES:
            return 'aws'
        return self.name.lower()`

and then: provider_name = provider.to_shorthand_string()

WDYT?

cloudrail/knowledge/context/gcp/gcp_resource.py
Tomerva: https://indeni.atlassian.net/browse/CR-2614

cloudrail/knowledge/rules/rule_metadata.py
alonashkenazi: looks better. Just add tests

PR #107, opened 2021-07-07
s3 bucket resource updated for cfn support

idan-aviv:


cloudrail/knowledge/context/aws/s3/s3_bucket.py
```diff
@@ -82,3 +86,16 @@ def is_tagable(self) -> bool:
     @property
     def is_public(self):
         return len(self.publicly_allowing_resources) > 0
+
+    def get_attribute(self, cfn_attribute_name: str):
```
yi2020, line 90: @idan-aviv @Viperain @alonashkenazi 
I don't like having this function in the context, it makes the context objects messy. The context should not care what is the source of the data (TF, CFN, Pulumi, live env).

yi2020 (changes requested):

cloudrail/knowledge/context/aws/s3/s3_bucket.py
idan-aviv: moved to cloudrail repository responsibility

yi2020 (approved):

PR #115, opened 2021-07-13
CR-2445: Need to modify logic for elasticsearch encryption at rest rule

Imanuel-Mizrahi:


cloudrail/knowledge/rules/aws/non_context_aware/encryption_enforcement_rules/encrypt_at_rest/es_encrypt_at_rest_rule.py
```diff
@@ -25,3 +26,8 @@ def execute(self, env_context: AwsEnvironmentContext, parameters: Dict[Parameter
 
     def should_run_rule(self, environment_context: AwsEnvironmentContext) -> bool:
         return bool(environment_context.elastic_search_domains)
+
+    @staticmethod
+    def _supported_es_domain_for_encryption(es_domain: ElasticSearchDomain) -> bool:
+        return (es_domain.es_domain_cluster_instance_type.split('.')[0] not in ('m3', 'r3', 't2')
+                and float(es_domain.es_domain_version) >= 5.1)
```
yi2020, line 33: What if the version has some character in it or some other issue in the future? Rule will crash.

yi2020 (changes requested):

cloudrail/knowledge/rules/aws/non_context_aware/encryption_enforcement_rules/encrypt_at_rest/es_encrypt_at_rest_rule.py
Imanuel-Mizrahi: Good point.
I found that we have actually only 2 versions which do not support es encryption (the list of versions are here: https://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/aes-supported-es-operations.html)
So, I switched to a static check against a set of these 2 versions

yi2020 (approved):

Tomerva (approved):

PR #122, opened 2021-07-19
CR-2466: not car azure ensure mysql servers enforce ssl

Imanuel-Mizrahi:


cloudrail/knowledge/context/azure/databases/azure_mysql_server.py
```diff
@@ -0,0 +1,40 @@
+from typing import Optional, List
+from cloudrail.knowledge.context.azure.azure_resource import AzureResource
+from cloudrail.knowledge.context.azure.constants.azure_resource_type import AzureResourceType
+
+
+class AzureMySqlServer(AzureResource):
```
yi2020, line 6: @Imanuel-Mizrahi you need to add this class to the documentation README files so they get pulled in.

yi2020 (changes requested):

cloudrail/knowledge/context/azure/databases/azure_mysql_server.py
Imanuel-Mizrahi: Thank you @yi2020 , added.

Tomerva (approved):

yi2020 (approved):

PR #123, opened 2021-07-19
add az aws resource doc

idan-aviv:


cloudrail/knowledge/context/aws/ec2/availability_zone.py
```diff
@@ -6,6 +6,11 @@
 
 class AvailabilityZone(AwsResource):
 
+    """
```
yi2020, line 9: Also need to add this file to the README's in the context documentation

yi2020 (changes requested):

cloudrail/knowledge/context/aws/ec2/availability_zone.py
idan-aviv: are you talking about i.e. docs/context/aws/<aws-resource>/README.md?
if so i see it is related to TF only, az aws resource fetched only from real environment

yi2020: The documentation is for the entire context model - the classes and their fields - it doesn't matter if they are built from TF, live or CFN.

docs/context/aws/availability_zone/README.md
```diff
@@ -0,0 +1,5 @@
+## ::: cloudrail.knowledge.context.aws.ec2.availability_zone.py
```
yi2020, line 1: Wrong path. Should be aws/ec2

PR #126, opened 2021-07-21
CR-2509: not car azure ensure sql servers audit enabled

Imanuel-Mizrahi:


cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class EnsureSqlServerAuditEnabledRule(AzureBaseRule):
+
+    def get_id(self) -> str:
+        return 'non_car_sql_servers_auditing_enabled'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for sql_server in env_context.sql_servers.values():
```
yi2020, line 16: Why do we need .values()?

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
```diff
@@ -0,0 +1,28 @@
...
+            if not sql_server.extended_auditing_policy \
+                or not sql_server.extended_auditing_policy.log_monitoring_enabled \
+                    or (0 < sql_server.extended_auditing_policy.retention_in_days <= 90):
+                issues.append(
+                    Issue(
+                        f'The {sql_server.get_type()} `{sql_server.get_friendly_name()}` does not have auditing enabled',
```
yi2020, line 22: Note that this issue will be triggered for retention_in_days being between 0 and 90, but the evidence doesn't make that clear.

yi2020 (changes requested):

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
Imanuel-Mizrahi: I used this logic based on the remediation steps mentioned in the rule_metadata for this rule.
@yi2020 ,perhaps we need to have different evidence strings? One for no audit enabled, and second if the audit enabled, but the retention is set to between 0 and 90?

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
Imanuel-Mizrahi: I do not think we need, but this is the same way already implemented for other rule, I will just fix both

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
yi2020: Yeah - split the if's and generate two string:

If it's not configured at all:
```
                        f'The {sql_server.get_type()} `{sql_server.get_friendly_name()}` does not have auditing enabled',
```

If not enough days:
```
                        f'The {sql_server.get_type()} `{sql_server.get_friendly_name()}` has auditing enabled, but for less than 90 days',
```

cloudrail/knowledge/context/azure/azure_environment_context.py
```diff
@@ -51,3 +53,4 @@ def __init__(self,
         self.security_center_auto_provisioning: AliasesDict[AzureSecurityCenterAutoProvisioning] = security_center_auto_provisioning or AliasesDict()
         self.security_center_contacts: AliasesDict[AzureSecurityCenterContact] = security_center_contacts or AliasesDict()
         self.vnet_gateways: AliasesDict[AzureVirtualNetworkGateway] = vnet_gateways or AliasesDict()
+        self.sql_server_extended_audit_policies: AliasesDict[AzureSqlServerExtendedAuditingPolicy] = sql_server_extended_audit_policies or AliasesDict()
```
alonashkenazi, line 56: @Imanuel-Mizrahi rename to mysql?

cloudrail/knowledge/context/azure/databases/azure_mssql_server_extended_auditing_policy.py
```diff
@@ -0,0 +1,43 @@
+from typing import List, Optional
+
+from cloudrail.knowledge.context.azure.azure_resource import AzureResource
+from cloudrail.knowledge.context.azure.constants.azure_resource_type import \
+    AzureResourceType
+
+
+class AzureSqlServerExtendedAuditingPolicy(AzureResource):
```
alonashkenazi, line 8: I think that you should mark this resource as not standalone

cloudrail/knowledge/context/azure/databases/azure_mssql_server_extended_auditing_policy.py
```diff
@@ -0,0 +1,43 @@
...
+        pass
+
+    def get_cloud_resource_url(self) -> Optional[str]:
+        return f'https://portal.azure.com/#@{self.tenant_id}/resource{self.server_id}/serverAuditing'
+
+    def get_friendly_name(self) -> str:
+        return f'Extended Audit policy for {self.server_id}'
+
+    def get_type(self, is_plural: bool = False) -> str:
+        if not is_plural:
```
alonashkenazi, line 36: retrun 'SQL Server Extended Auditing ' +  'Policy'  if not is_plural else 'Policies'

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
```diff
@@ -0,0 +1,34 @@
...
+
+class EnsureSqlServerAuditEnabledRule(AzureBaseRule):
+
+    def get_id(self) -> str:
+        return 'non_car_sql_servers_auditing_enabled'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for sql_server in env_context.sql_servers:
+            if not sql_server.extended_auditing_policy.log_monitoring_enabled \
```
alonashkenazi, line 17: Can the sql_server extended_auditing_policy be none?
should you add to your check:
if sql_server.extended_auditing_policy is not None?

Imanuel-Mizrahi: No, the current builders will always fill in this field.

cloudrail/knowledge/context/azure/azure_environment_context.py
Imanuel-Mizrahi: no, mysql is a different resource in Azure, this should be only sql_server.
This is also how the sql_servers are named as well.

cloudrail/knowledge/context/azure/databases/azure_mssql_server_extended_auditing_policy.py
Imanuel-Mizrahi: makes sense - changed it

cloudrail/knowledge/context/azure/databases/azure_mssql_server_extended_auditing_policy.py
Imanuel-Mizrahi: nice one, done.

cloudrail/knowledge/context/azure/databases/azure_sql_server.py
```diff
@@ -23,7 +28,7 @@ def get_name(self) -> str:
         return self.server_name
 
     def get_cloud_resource_url(self) -> Optional[str]:
-        pass
+        return f'https://portal.azure.com/#@{self.tenant_id}/resource{self._id}/overview'
```
Tomerva, line 31: i think you missed a slash

Tomerva (approved):

cloudrail/knowledge/context/azure/databases/azure_sql_server.py
Imanuel-Mizrahi: Didn't missed it, the ID already includes the slash.
There is a context test for the "get_cloud_resource", and it's fine.

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
yi2020: The order here doesn't make sense. You do test for none, but after you already tried using extended_auditing_policy. This will fail.

yi2020 (changes requested):

cloudrail/knowledge/rules/azure/non_context_aware/ensure_sql_server_audit_enabled_rule.py
Imanuel-Mizrahi: Fixed order

yi2020 (approved):

PR #141, opened 2021-07-28
CR-2496: [Not CAR] [Azure] Ensure Storage Account is allowing network access to trusted Azure services

Imanuel-Mizrahi:


cloudrail/knowledge/rules/azure/non_context_aware/storage_account_allow_network_access_trusted_azure_services_rule.py
```diff
@@ -0,0 +1,26 @@
...
+
+class StorageAccountAllowNetworkAccessTrustedAzureResourcesRule(AzureBaseRule):
+
+    def get_id(self) -> str:
+        return 'non_car_storage_account_network_access_trusted_azure_services'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for storage_account in env_context.storage_accounts:
+            if 'AzureServices' not in storage_account.network_rules.bypass_traffic:
```
alonashkenazi, line 17: should we have enum that AzureServices is a member there?

alonashkenazi (approved):

cloudrail/knowledge/rules/azure/non_context_aware/storage_account_allow_network_access_trusted_azure_services_rule.py
Imanuel-Mizrahi: @alonashkenazi , this is some general question when using ENUM, and when not.
As far as I understand, we need to use ENUM, if we know that the values will not be changed / modified in the future, otherwise, not worth to mess with it.
Currently, there is some set of options available for bypass, but as we cannot be sure if there will be something else in the future, I would leave it like this, and not use an ENUM.

yi2020 (approved):

Viperain (approved):

cloudrail/knowledge/rules/azure/non_context_aware/storage_account_allow_network_access_trusted_azure_services_rule.py
Imanuel-Mizrahi: Modified with Enum usage

yi2020 (approved):

PR #144, opened 2021-08-04
RemediationSteps: add new rule metadata class

idan-aviv:


cloudrail/knowledge/rules/rule_metadata.py
```diff
@@ -60,15 +60,21 @@ class ResourceType(str, Enum):
...
+
+
 @dataclass
 class RuleMetadata:
     rule_id: str
     name: str
     description: str
     logic: str
     severity: RuleSeverity
-    iac_remediation_steps: str
```
yi2020, line 70: 1. This is breaking backwards compatibility - if someone wrote a custom rule it no longer will work. (@alonashkenazi )
2. All existing rules need to be updated to use the new format.

yi2020 (changes requested):

cloudrail/knowledge/rules/rule_metadata.py
alonashkenazi: @yi2020 The ctx has a lot of problems, we will need to refactor it. At this point, since we don't have customers that are using the ctx (like external API) I prefer to keep and improve the ctx object modeling and take the "risk"

idan-aviv: @yi2020 if no additional comments please approve, thanks

yi2020 (approved):

PR #145, opened 2021-08-08
Feature/cr 2263 non car diagnostic logs enabled in app services

lirshindalman:


cloudrail/knowledge/context/azure/webapp/diagnostic_logs.py
```diff
@@ -0,0 +1,8 @@
+
+class DiagnosticLogs:
+
+    def __init__(self, detailed_error_logging_enabled: bool, http_logging_enabled: bool, request_tracing_enabled: bool) -> None:
```
alonashkenazi, line 4: __init__ shouldn't have a return type. remove the -> None

cloudrail/knowledge/context/azure/webapp/azure_app_service_config.py
```diff
@@ -17,13 +18,14 @@ class AzureAppServiceConfig(AzureResource):
 
     """
 
-    def __init__(self, name, ftps_state: FtpsState, auth_settings: AuthSettings, minimum_tls_version: str, http2_enabled: bool) -> None:
+    def __init__(self, name, ftps_state: FtpsState, auth_settings: AuthSettings, minimum_tls_version: str, http2_enabled: bool, logs: DiagnosticLogs) -> None:
```
alonashkenazi, line 21: __init__ shouldn't have a return type. remove the -> None

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
```diff
@@ -0,0 +1,34 @@
...
+
+class AppServiceDiagnosticLogsRule(AzureBaseRule):
+    def get_id(self) -> str:
+        return 'non_car_diagnostic_logs_enabled_in_app_services'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+
+        for app_service in env_context.app_services:
+            logs: List[str] = []
```
alonashkenazi, line 18: rename to evidance

tests/knowledge/rules/azure/non_context_aware/test_app_service_non_car_diagnostic_logs_enabled_in_app_services.py
```diff
@@ -0,0 +1,47 @@
...
+        app_service_config.logs = logs
+        app_service.app_service_config = app_service_config
+        context = AzureEnvironmentContext(app_services=AliasesDict(app_service))
+
+        # Act
+        result = self.rule.run(context, {})
+        # Assert
+        if should_alert:
+            self.assertEqual(RuleResultType.FAILED, result.status)
+            self.assertEqual(1, len(result.issues))
```
alonashkenazi, line 45: I suggest that in your case you will also assert on the evidence string

cloudrail/knowledge/context/azure/webapp/azure_app_service_config.py
```diff
@@ -17,13 +18,14 @@ class AzureAppServiceConfig(AzureResource):
```
Tomerva, line 18: missing documentation of the new attribute

Tomerva:
need to update docs/context/azure/webapp/README.md to include the new rule

cloudrail/knowledge/context/azure/webapp/diagnostic_logs.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class DiagnosticLogs:
```
alonashkenazi, line 5: @lirshindalman you need to add documentation for this as well

alonashkenazi (changes requested):

cloudrail/knowledge/rules/azure/azure_rules_metadata.yaml
```diff
@@ -1,7 +1,7 @@
 templates:
```
Tomerva, line 1: why did you remove the brackets ( < > ) ? the links in the metadata yaml should be surrounded by them

cloudrail/knowledge/context/azure/webapp/azure_app_service_config.py
```diff
@@ -14,16 +15,18 @@ class AzureAppServiceConfig(AzureResource):
             auth_settings: App service authentication settings.
             minimum_tls_version: The minimum supported TLS version for the function app.
             http2_enabled: Indication if http2 protocol should be enabled or not.
+            logs: The DiagnosticLogs indicate if the logs (detailed error messages, HTTP logging, and failed requests tracing) are  enabled or not
```
yi2020, line 18: extra space in `are  enabled`

cloudrail/knowledge/context/azure/webapp/diagnostic_logs.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
+
+
+
+@dataclass
+class DiagnosticLogs:
+    """
+            Attributes:
+                detailed_error_logging_enabled: Indicate if the detailed error logging enable or disable.
```
yi2020, line 9: `is enabled or disabled`, same for the others.

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
```diff
@@ -0,0 +1,34 @@
...
+class AppServiceDiagnosticLogsRule(AzureBaseRule):
+    def get_id(self) -> str:
+        return 'non_car_diagnostic_logs_enabled_in_app_services'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+
+        for app_service in env_context.app_services:
+            evidence: List[str] = []
+            if app_service.app_service_config is not None and app_service.app_service_config.logs is not None:
```
yi2020, line 18: what if .logs is None?

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
```diff
@@ -0,0 +1,34 @@
...
+                        f'The web app `{app_service.get_friendly_name()}` does not have HTTP logging enabled')
+                if not app_service.app_service_config.logs.request_tracing_enabled:
+                    evidence.append(
+                        f'The web app `{app_service.get_friendly_name()}` does not have request tracing enabled')
+                if not app_service.app_service_config.logs.detailed_error_logging_enabled:
+                    evidence.append(
+                        f'The web app `{app_service.get_friendly_name()}` does not have detailed error logging enabled')
+            if evidence:
+                issues.append(
+                    Issue('\n'.join(evidence), app_service, app_service))
```
yi2020, line 30: Did you try to see how this output would look like in the CLI and the web UI? We don't normally join evidences using a newline character.

cloudrail/knowledge/context/azure/webapp/diagnostic_logs.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
+
+
+
+@dataclass
+class DiagnosticLogs:
```
yi2020, line 6: In order for this class to be in the READMEs, you need to add a reference to. this Python package in the relevant location in the readmes.

yi2020 (changes requested):

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
lirshindalman: I handle it in the builder, if there are no logs the default is False and then the issue will be with the 3 cases

yi2020: That doesn't make sense, you're "abusing" the type-unsafety of Python. If you declare that this field needs to have a certain object in it, then it can either have that object or None. It shouldn't have False (even if Python supports it). @alonashkenazi @ori-bar  we should generally think about increasing type-safety.

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
```diff
@@ -0,0 +1,35 @@
...
+                        f'The web app `{app_service_name}` does not have HTTP logging enabled')
+                if not app_service.app_service_config.logs.request_tracing_enabled:
+                    evidence.append(
+                        f'The web app `{app_service_name}` does not have request tracing enabled')
+                if not app_service.app_service_config.logs.detailed_error_logging_enabled:
+                    evidence.append(
+                        f'The web app `{app_service_name}` does not have detailed error logging enabled')
+            if evidence:
+                issues.append(
+                    Issue(', '.join(evidence), app_service, app_service))
```
yi2020, line 31: Use a period instead of a comma for joining. The reason is that you have a capital T at the beginning of each evidence, and there shouldn't be a capital letter following a comma in English.

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
yi2020: I'm sorry @lirshindalman I misunderstood your comment. I now looked at how you do it in the builders:
```
        Scanner:
                return AzureAppServiceConfig(name=attributes['name'],
                                     ftps_state=FtpsState(attributes['properties']['ftpsState']),
                                     auth_settings=AuthSettings(attributes['properties']['siteAuthEnabled']),
                                     minimum_tls_version=attributes['properties']['minTlsVersion'],
                                     http2_enabled=attributes['properties']['http20Enabled'])
                                     http2_enabled=attributes['properties']['http20Enabled'],
                                     logs=DiagnosticLogs(detailed_error_logging_enabled,
                                                         http_logging_enabled,
                                                         request_tracing_enabled))

        Terraform:
        logs: Optional[DiagnosticLogs] = DiagnosticLogs(False, False, False)
        if logs_dict := self._get_known_value(attributes, 'logs'):
            logs.detailed_error_logging_enabled = self._get_known_value(logs_dict[0], 'detailed_error_messages_enabled', False)
            logs.request_tracing_enabled = self._get_known_value(logs_dict[0], 'failed_request_tracing_enabled', False)
            logs.http_logging_enabled = True if self._is_known_value(logs_dict[0], 'http_logs') else False
```

Which means the field will not be empty with these builders. However, there are other infrastructure-as-code languages in the world, even for Azure. We plan on supporting those at some point - like Pulumi, Bicep, ARM. If the developer who writes the builders for those provides a None for this field, the rule will crash. Unfortunately, this has happened many times, even last week, and so we need to build the rules in a way where they're None-proof. So my comment still stands.

Tomerva (approved):

alonashkenazi (approved):

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
```diff
@@ -0,0 +1,38 @@
...
+        return 'non_car_diagnostic_logs_enabled_in_app_services'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+
+        for app_service in env_context.app_services:
+            evidence: List[str] = []
+            if app_service.app_service_config is not None:
+                if app_service.app_service_config.logs is None:
+                    app_service.app_service_config['logs'] = DiagnosticLogs(False, False, False)
```
yi2020, line 21: Rules are not allowed to modify the content within a a context object's class.

yi2020 (changes requested):

cloudrail/knowledge/rules/azure/non_context_aware/app_service_non_car_diagnostic_logs_enabled_in_app_services_rule.py
yi2020: The solution here would be to add this in the case of None:
```
f'The web app `{app_service_name}` does not have logging enabled')
```

Make sure not to run the other if's if the `.logs` is None, as the rule will crash if you do.

yi2020 (approved):

PR #149, opened 2021-08-08
Added rule KubernetesClusterRbacEnabledRule

Tomerva:


alonashkenazi (approved):

cloudrail/knowledge/context/azure/aks/azure_kubernetes_cluster.py
```diff
@@ -0,0 +1,22 @@
...
+
+from cloudrail.knowledge.context.azure.azure_resource import AzureResource
+from cloudrail.knowledge.context.azure.constants.azure_resource_type import AzureResourceType
+
+
+class AzureKubernetesCluster(AzureResource):
+    def __init__(self, name: str, enable_rbac: bool):
+        super().__init__(AzureResourceType.AZURERM_KUBERNETES_CLUSTER)
+        self.name: str = name
+        self.enable_rbac: bool = enable_rbac
```
alonashkenazi, line 11: Something to think about:
the name that you gave to the prop is can be misleading as there is azure_rbac_enabled prop as well in tf, maybe call it 
  role_based_access_control_enable? 
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#azure_rbac_enabled

The prop in the scanner is enableRBAC so this I don't know what is the right answer...

Tomerva: i named it according to scanner data.. also role_based_access_control_enable is practically the same as rbac_enable so i dont see a reason to change it

mkdocs.yml
```diff
@@ -64,6 +64,7 @@ nav:
       - 'security': 'context/azure/security/README.md'
       - 'keyvault': 'context/azure/keyvault/README.md'
       - 'storage': 'context/azure/storage/README.md'
+      - 'aks': 'context/azure/aks/README.md'
```
yi2020, line 67: Can you add these entries sorted A-Z? I realize Azure is messy here, but just noticed it due to this one starting with a.

yi2020 (changes requested):

yi2020 (approved):

PR #152, opened 2021-08-09
CR-2337: [Not CAR] [Azure] Ensure unattached Managed Disks are encrypted

Imanuel-Mizrahi:
Modified resource, builders, tests, rule ready

### DIFF
--- a/cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
+++ b/cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
@@ -1,0 +1,60 @@
+from enum import Enum
+from typing import Optional, List
+
+from cloudrail.knowledge.context.azure.azure_resource import AzureResource
+from cloudrail.knowledge.context.azure.constants.azure_resource_type import AzureResourceType
+
+class StorageAccountType(Enum):
+    STANDARD_LRS = 'Standard_LRS'
+    PREMIUM_LRS = 'Premium_LRS'
+    STANDARDSSD_LRS = 'StandardSSD_LRS'
+    ULTRASSD_LRS = 'UltraSSD_LRS'
+
+
+class ManagedDiskCreateOption(Enum):
+    IMPORT = 'Import'
+    EMPTY = 'Empty'
+    COPY = 'Copy'
+    FROMIMAGE = 'FromImage'
+    RESTORE = 'Restore'
+
+
+class AzureManagedDisk(AzureResource):
+    """
+        Attributes:
+            name: The Managed Disk name
+            storage_account_type: The type of storage used for the Managed Disk
+            create_option: The method used to create the Managed Disk
+            disk_encryption_set_id: The ID of the Disk Encryption Set which is being used to encrypt the Managed Disk if any
+            disk_encryption_enabled: Indication if the Managed Disk is encrypted using platform key.
+    """
+
+    def __init__(self, name: str, storage_account_type: StorageAccountType,
+                 create_option: ManagedDiskCreateOption, disk_encryption_set_id: Optional[str], disk_encryption_enabled: bool):
+        super().__init__(AzureResourceType.AZURERM_MANAGED_DISK)
+        self.name: str = name
+        self.with_aliases(name)
+        self.storage_account_type: StorageAccountType = storage_account_type
+        self.create_option: ManagedDiskCreateOption = create_option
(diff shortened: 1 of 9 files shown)


cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__(AzureResourceType.AZURERM_MANAGED_DISK)
+        self.name: str = name
+        self.with_aliases(name)
+        self.storage_account_type: StorageAccountType = storage_account_type
+        self.create_option: ManagedDiskCreateOption = create_option
+        self.disk_encryption_set_id: Optional[str] = disk_encryption_set_id
+        self.disk_encryption_enabled: bool = disk_encryption_enabled
+
+    def get_cloud_resource_url(self) -> Optional[str]:
+        return f'https://portal.azure.com/#@{self.tenant_id}/resource{self._id}/overview'
```
Tomerva, line 43: missing / between resource and {self._id} ?

cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
```diff
@@ -0,0 +1,60 @@
...
+            create_option: The method used to create the Managed Disk
+            disk_encryption_set_id: The ID of the Disk Encryption Set which is being used to encrypt the Managed Disk if any
+            disk_encryption_enabled: Indication if the Managed Disk is encrypted using platform key.
+    """
+
+    def __init__(self, name: str, storage_account_type: StorageAccountType,
+                 create_option: ManagedDiskCreateOption, disk_encryption_set_id: Optional[str], disk_encryption_enabled: bool):
+        super().__init__(AzureResourceType.AZURERM_MANAGED_DISK)
+        self.name: str = name
+        self.with_aliases(name)
```
Tomerva, line 36: is name unique across accounts/regions? if not, then dont use it as alias

cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
Imanuel-Mizrahi: @Tomerva , this is not missing.
The ID include / at the beginning, as follows:
"/subscriptions/230613d8-3b34-4790-b650-36f31045f19a/resourceGroups/CR23371-RG/providers/Microsoft.Compute/disks/cr23371-disk"
So adding another / is not good.

cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def get_type(self, is_plural: bool = False) -> str:
+        return 'Managed ' + 'Disk' if not is_plural else 'Disks'
+
+    def get_name(self) -> str:
+        return self.name
+
+    @property
+    def is_encrypted(self) -> bool:
+        return self.disk_encryption_enabled or self.disk_encryption_set_id
```
Tomerva, line 60: this is a bit odd, because if you have the set_id then the encryption should be enabled right?

if so, then maybe in the init you can override disk_encryption_enabled if disk_encryption_set_id is not none ?

Imanuel-Mizrahi: @Tomerva , I had some thoughts about it.
The point is that for this resource, we have 2 possible ways to set encryption (similar to KMS key in AWS: customer encrypted or AWS encrypted), since it is considered as 2 separate options, I left it separate like this.

cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
Imanuel-Mizrahi: OK, good point

yi2020 (approved):

cloudrail/knowledge/context/azure/disk/azure_managed_disk.py
Tomerva: the question is, does it line up with scanner data?

like, if you dont set disk_encryption_enabled but disk_encryption_set_id is set, will disk_encryption_enabled be false? if not, then its not really synced..

Imanuel-Mizrahi: @Tomerva  In such state, disk_encryption_enabled will be None (the field will be completely missing from scanner data)

Tomerva (approved):

PR #161, opened 2021-08-17
move compliance to new format

Viperain:


alonashkenazi (approved):
@Viperain you should let @ori-bar know about this change as she is touching this file

PR #166, opened 2021-08-22
fix rule to not contains state  of data

Viperain:


cloudrail/knowledge/context/cloud_provider.py
```diff
@@ -18,4 +18,4 @@ def from_string(string):
     def to_shorthand_string(self):
         if self == CloudProvider.AMAZON_WEB_SERVICES:
             return 'aws'
-        return self.name.lower()
+        return str(self.name).lower()
```
ori-bar, line 21: why?

ori-bar (approved):

cloudrail/knowledge/context/cloud_provider.py
Viperain: lint

PR #171, opened 2021-08-24
CR-3042: Drift detection: Adding a new tag to an S3 bucket is not detected

Imanuel-Mizrahi:


cloudrail/knowledge/context/aws/iam/iam_instance_profile.py
```diff
@@ -34,12 +34,12 @@ def get_type(self, is_plural: bool = False) -> str:
             return 'IAM Instance profiles'
 
     def get_arn(self) -> str:
-        pass
+        return f'arn:aws:iam::{self.account}:instance-profile/{self.iam_instance_profile_name}'
 
     def get_cloud_resource_url(self) -> str:
-        return '{0}ec2/v2/home?region={1}#InstanceDetails:instanceId={2}'\
```
yi2020, line 40: Why did you change this?

Imanuel-Mizrahi: @yi2020 , you are fast, I didn't finished my work on this branch...
Anyway, this once I had to change, since the link is based on the data we get from the EC2 instance.
The problem is: you can create IAM instance profile, which won't be yet attached to any EC2 (as we have in one of our test scenarios), in such case, this link is not relevant at all.
Thus, I changed the link to use the IAM role itself, and I think this is more accurate.

yi2020:
Ah makes sense

Tomerva (approved):

yi2020 (approved):

PR #172, opened 2021-08-24
Added rule FsxWindowsFileSystemEncryptedAtRestWithCustomerManagedCmkRule

Tomerva:


cloudrail/knowledge/context/aws/fsx/fsx_windows_file_system.py
```diff
@@ -0,0 +1,44 @@
...
+        self.kms_data: Optional[KmsKey] = None
+        self.with_aliases(fsx_windows_file_system_id)
+
+    def get_keys(self) -> List[str]:
+        return [self.fsx_windows_file_system_id]
+
+    def get_type(self, is_plural: bool = False) -> str:
+        return 'FSx Windows File System' + 's' if is_plural else ''
+
+    def get_arn(self) -> str:
```
Imanuel-Mizrahi, line 33: @Tomerva , if the resource is tagable, we need to fill in some ARN.
It's because the tags are being filled to each resource, while matching the ARN and the tags list from AWS.

Imanuel-Mizrahi (changes requested):

Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/aws/fsx/fsx_windows_file_system.py
yi2020: @Imanuel-Mizrahi @Tomerva @Viperain is there a way to add this logic to a test, so resources aren't added with this issue?

yi2020 (approved):

cloudrail/knowledge/context/aws/fsx/fsx_windows_file_system.py
Imanuel-Mizrahi: While working on tags for drift, I found that we actually have 2 ways to get the "tags" for AWS resource **scanner**.
One will be using the ARN, the other will try to get "tags" from the API call of the resource itself, if there is one.
So, to be honest, we do not need always the "get_arn" for a "tagable" resource, but only if the "tags" does not appear in the same API call which we build the resource with.
So, I guess we still need the programmer to manually check it.

PR #176, opened 2021-08-26
CR-3090: add-binary-check

olozz:


.github/workflows/white-list-binary.txt
```diff
@@ -0,0 +1,3 @@
+---exclusions for binary files---
+account-data.zip$
+account-data-for-drift.zip$
```
ivantikal, line 3: white list should be empty here (we don't want to add zip files here)

ivantikal (approved):

PR #190, opened 2021-09-01
IacType: relocate from cloudrail repo

idan-aviv:


Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/mergeable.py
```diff
@@ -82,13 +84,18 @@ def origin(self) -> EntityOrigin:
         if self.is_pseudo:
             return EntityOrigin.PSEUDO
         elif self.iac_state:
-            return EntityOrigin.TERRAFORM
+            if self.iac_state.iac_type == IacType.TERRAFORM:
+                return EntityOrigin.TERRAFORM
+            elif self.iac_state.iac_type == IacType.CLOUDFORMATION:
+                return EntityOrigin.CLOUDFORMATION
+            else:
+                return EntityOrigin.NONE
```
yi2020, line 92: Do we know what to do with NONE? Wouldn't it make more sense to throw an exception here?

Tomerva: i agree, None makes no sense and if we get None here for some reason we should raise an exception

idan-aviv (commented):
@Tomerva @yi2020 done

cloudrail/knowledge/context/mergeable.py
idan-aviv: @yi2020 @Tomerva done

idan-aviv:
duplicated

PR #195, opened 2021-09-05
Feature/cr 3134 aws sqs queue policy modification

lirshindalman:


cloudrail/knowledge/context/aws/resources/apigateway/rest_api_gw.py
```diff
@@ -44,6 +44,8 @@ def __init__(self,
         self.is_public: bool = None
         self.api_gateway_methods: List[ApiGatewayMethod] = []
         self.agw_methods_with_valid_integrations_and_allowed_lambda_access: List[ApiGatewayMethod] = []
+        self.resource_based_policy: RestApiGwPolicy = None
```
Imanuel-Mizrahi, line 47: Why we only add "resource_based_policy" just for this resource, I think we need to add it to all the other resources, no?
In addition, perhaps we can simply add to the "ResourceBasedPolicy" class, the attribute: "resource_based_policy", so all who inherits from it, will already have it, right?

Imanuel-Mizrahi (changes requested):

Imanuel-Mizrahi:
@lirshindalman , a general recommendation about this PR.
As this change affect many resources, I would say that it is actually better to create a beta version, and make sure all server tests pass, before we proceed to merge this PR.
We might have some unexpected surprises...

cloudrail/knowledge/context/aws/resources/apigateway/rest_api_gw.py
lirshindalman: but each resource has a different resource_based_policy. example LambdaPolicy or S3Policy

cloudrail/knowledge/context/aws/aws_connection_builder.py
```diff
@@ -783,7 +782,7 @@ def _evaluate_private_permissions_connection(cls, aws_client: AwsClient,
                     policy_evaluation_result = role.policy_evaluation_result_map[target_resource.get_arn()]
                 else:
                     resource_based_policies: List[Policy] = [target_resource.resource_based_policy] \
-                        if isinstance(target_resource, ResourceBasedPolicy) and target_resource.resource_based_policy \
+                        if hasattr(target_resource, 'resource_based_policy') and target_resource.resource_based_policy \
```
Tomerva, line 785: hasattr is a very "hacky" solution, in most cases it should be avoided and should be solved by proper OOP design, like inheritance as discussed

Tomerva: i see you added class AwsResourceWithBasedPolicy. so instead of hasattr you should check if isinstance(target_resource, AwsResourceWithBasedPolicy)

cloudrail/knowledge/context/aws/resources/apigateway/rest_api_gw.py
```diff
@@ -17,7 +17,7 @@ class ApiGatewayType(Enum):
     PRIVATE = 'PRIVATE'
 
 
-class RestApiGw(ResourceBasedPolicy):
+class RestApiGw(AwsResource):
```
Tomerva, line 20: should inherit from AwsResourceWithBasedPolicy

cloudrail/knowledge/context/aws/resources/aws_resource_with_based_policy.py
```diff
@@ -0,0 +1,20 @@
+from abc import abstractmethod
+from typing import Optional, List
+from cloudrail.knowledge.context.aws.resources.aws_resource import AwsResource
+from cloudrail.knowledge.context.aws.resources.resource_based_policy import ResourceBasedPolicy
+from cloudrail.knowledge.context.aws.resources.service_name import AwsServiceName, AwsServiceAttributes
+
+
+class AwsResourceWithBasedPolicy(AwsResource):
```
Tomerva, line 8: need to figure out a better name for this class.. i think PoliciedResource is ok, WDYT?

cloudrail/knowledge/context/aws/resources/lambda_/lambda_function.py
```diff
@@ -1,19 +1,18 @@
...
 from cloudrail.knowledge.context.aws.resources.networking_config.network_configuration import NetworkConfiguration
 from cloudrail.knowledge.context.aws.resources.networking_config.network_entity import NetworkEntity
-from cloudrail.knowledge.context.aws.resources.resource_based_policy import ResourceBasedPolicy
 from cloudrail.knowledge.context.aws.resources.service_name import AwsServiceAttributes, AwsServiceName, AwsServiceType
 from cloudrail.knowledge.utils.arn_utils import are_arns_intersected, is_valid_arn
 
 
-class LambdaFunction(NetworkEntity, ResourceBasedPolicy, AwsClient):
+
+class LambdaFunction(NetworkEntity, AwsResource, AwsClient):
```
Tomerva, line 15: should also inherit from your new class, like every resource that has resource_based_policy attribute. please search for all occurrences of resource_based_policy

Tomerva (changes requested):

cloudrail/knowledge/context/aws/resources/apigateway/rest_api_gw.py
lirshindalman: inherit from PoliciedResource

cloudrail/knowledge/context/aws/resources/es/elastic_search_domain.py
```diff
@@ -39,7 +39,7 @@ class ElasticSearchDomain(NetworkEntity, INetworkConfiguration):
             is_in_vpc: True if the ElasticSearch Domain is accessible at a specific
                 VPC.
             ports: The ports the ElasticSearch is listening on.
-            policy: The resource policy used with the domain.
+            resource_based_policy: The resource policy used with the domain.
```
Imanuel-Mizrahi, line 42: I guess we can remove the "resource_based_policy" field here as well.

cloudrail/knowledge/context/aws/resources/lambda_/lambda_function.py
```diff
@@ -25,6 +23,7 @@ class LambdaFunction(NetworkEntity, ResourceBasedPolicy, AwsClient):
             vpc_config: The VPC configuration of the Lambda Function, if one was set.
             log_group: The matching log group associated with the Lambda Function.
             xray_tracing_enabled: Indication if X-Ray tracing is enabled for incoming requests.
+            resource_based_policy: The resource policy of the Lambda Function, if any is defined.
```
Imanuel-Mizrahi, line 26: Remove the "resource_based_policy" field

cloudrail/knowledge/context/aws/resources/secretsmanager/secrets_manager_secret.py
```diff
@@ -1,17 +1,16 @@
...
 
 
-class SecretsManagerSecret(AwsResource):
+class SecretsManagerSecret(PoliciedResource):
     """
         Attributes:
             sm_name: The name of the owning SageManager.
             arn: The ARN of this secret.
-            policy: The resource policy to use with this secret, if any.
+            resource_based_policy: The resource policy to use with this secret, if any.
```
Imanuel-Mizrahi, line 13: Remove the "resource_based_policy" field

Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/aws/resources/aws_policied_resource.py
```diff
@@ -0,0 +1,26 @@
+from abc import abstractmethod
+from typing import List
+from cloudrail.knowledge.context.aws.resources.aws_resource import AwsResource
+from cloudrail.knowledge.context.aws.resources.resource_based_policy import ResourceBasedPolicy
+from cloudrail.knowledge.context.aws.resources.service_name import AwsServiceName, AwsServiceAttributes
+
+
+class PoliciedResource(AwsResource):
+    """
+    This class is the parent of all resources that have a based policy.
```
Tomerva, line 10: that have a **resource** based policy

cloudrail/knowledge/context/aws/resources/es/elastic_search_domain.py
```diff
@@ -67,7 +67,8 @@ def __init__(self,
         self.encrypt_at_rest_state: bool = encrypt_at_rest_state
         self.encrypt_node_to_node_state: bool = encrypt_node_to_node_state
 
-        super().__init__(domain_name, account, region, AwsServiceName.AWS_ELASTIC_SEARCH_DOMAIN)
+        NetworkEntity.__init__(self, domain_name, account, region, AwsServiceName.AWS_ELASTIC_SEARCH_DOMAIN)
+        PoliciedResource.__init__(self, account, region, AwsServiceName.AWS_ELASTIC_SEARCH_DOMAIN)
```
Tomerva, line 71: these 2 lines basically do the same thing

read this https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance

cloudrail/knowledge/context/aws/resources/lambda_/lambda_function.py
```diff
@@ -34,8 +33,8 @@ def __init__(self, account: str, region: str, arn: str, qualified_arn: str, func
                  runtime: str, vpc_config: NetworkConfiguration, xray_tracing_enabled: bool):
         NetworkEntity.__init__(self, function_name, account, region, AwsServiceName.AWS_LAMBDA_FUNCTION,
                                AwsServiceAttributes(aws_service_type=AwsServiceType.LAMBDA.value, region=region))
-        ResourceBasedPolicy.__init__(self, account, region, AwsServiceName.AWS_LAMBDA_FUNCTION,
-                                     AwsServiceAttributes(aws_service_type=AwsServiceType.LAMBDA.value, region=region))
+        PoliciedResource.__init__(self, account, region, AwsServiceName.AWS_LAMBDA_FUNCTION,
```
Tomerva, line 36: same comment (https://stackoverflow.com/questions/3277367/how-does-pythons-super-work-with-multiple-inheritance)

cloudrail/knowledge/context/aws/resources/lambda_/lambda_function.py
```diff
@@ -78,7 +76,7 @@ def parse_qualifier_from_arn(qualified_arn: str) -> str:
             resource_parts: List[str] = arn_sections_dict['resource'].split(':')
             if len(resource_parts) == 3:
                 return resource_parts[-1]
-        return ''
+        return None
```
Tomerva, line 79: if it can return None, the return type should be Optional[str]

also please make sure all places that use this function, use it safely (IE, know that it can be None or it doesnt matter to them)

cloudrail/knowledge/rules/aws/non_context_aware/policy_wildcard_violation/ensure_policy_not_use_wildcard_actions_rules.py
```diff
@@ -35,7 +35,7 @@ def get_id(self) -> str:
 
     @staticmethod
     def _get_entity_policy(entity: CloudWatchLogsDestination) -> Policy:
```
Tomerva, line 37: this method is redundant here
because all these resources are of the same group (PoliciedResource) they all have resource_based_policy attribute.

so instead of implementing _get_entity_policy everywhere, you can either implement it on EnsurePolicyNotUseWildcardRules once, or (the more preferred way) change the return type of _get_rule_entities to `Optional[List[PoliciedResource]]` and then instead of doing `policy = self._get_entity_policy(entity)` you can just access it (`entity.resource_based_policy`) without needing to check the type with isinstance

Tomerva (changes requested):

cloudrail/knowledge/context/aws/resources/es/elastic_search_domain.py
lirshindalman: In this case where each class has its own positional arguments for its init, it butter to call each class's own init method.

Tomerva (approved):

PR #202, opened 2021-09-09
CR-3158: [AWS] RDS instance ID is not collected correctly for TF

Imanuel-Mizrahi:
Bugfix/cr 3158 fix rds instance id builder


cloudrail/knowledge/context/aws/resources/rds/rds_instance.py
```diff
@@ -21,7 +21,7 @@ class RdsInstance(NetworkEntity, INetworkConfiguration):
             security_group_ids: The IDs of the security groups in use
                 with the instance.
             db_cluster_id: The cluster ID, if this instance is part of a cluster,
-                or None otherwise.
+                or the DB instance ID otherwise.
```
Tomerva, line 24: this change seems forced.. cant you just change get_cloud_resource_url to use either cluster id or instance id as needed?

Imanuel-Mizrahi: @Tomerva ,
RDS instance could be a part of a cluster, or a standalone.
If it is a standalone instance, we will use the DB instance ID itself, otherwise, we will use the cluster ID, which will also be used for the resource URL.
Do you recommend creating 2 separate fields, one for each type of RDS instance, and then create a condition inside the get_cloud_resource_url function?

Tomerva: basically yes, one for cluster id one for instance id, and then as you said - create aa condition in get_cloud_resource_url

Imanuel-Mizrahi: @Tomerva , done.
Please review.
I will open a PR for server soon, with the modified tests for the new fields.

cloudrail/knowledge/context/aws/resources/rds/rds_instance.py
```diff
@@ -95,8 +99,12 @@ def get_type(self, is_plural: bool = False) -> str:
             return 'RDS Instances'
 
     def get_cloud_resource_url(self) -> Optional[str]:
-        return '{0}rds/home?region={1}#database:id={2};is-cluster=false'\
-            .format(self.AWS_CONSOLE_URL, self.region, self.db_cluster_id)
+        if self.db_cluster_id:
+            return '{0}rds/home?region={1}#database:id={2};is-cluster=false'\
```
Tomerva, line 103: is-cluster should be false here?

if the URL is the same in both cases (when using db_cluster_id and instance_id), then instead of the if do:

.....format(self.AWS_CONSOLE_URL, self.region, self.db_cluster_id or self.instance_id)

cloudrail/knowledge/context/azure/resources_builders/terraform/virtual_machine_scale_set_builder.py
```diff
@@ -45,6 +45,7 @@ def _build_vmss(attributes: dict, os_type: OperatingSystemType, vm_tf_type: str,
                                           bool(os_disk_profile[0].get('managed_disk_type'))),
                                                               data_disks_list)
     else:
+        os_disk_profile = attributes['os_disk']
```
Tomerva, line 48: why did you restore it? its still unused

cloudrail/knowledge/context/aws/resources/rds/rds_instance.py
Imanuel-Mizrahi: modified to "is-cluster=true"

cloudrail/knowledge/context/azure/resources_builders/terraform/virtual_machine_scale_set_builder.py
Imanuel-Mizrahi: some merging mix, removed it

Tomerva (approved):

PR #204, opened 2021-09-09
Feature/cr 3121 compliance support multiple cont

Viperain:


cloudrail/knowledge/rules/rules_metadata_store.py
```diff
@@ -123,8 +123,10 @@ def _parse_compliance(compliance_dict: dict) -> Dict[BenchmarkType, Dict[str, st
         result = {}
         for framework, versions in compliance_dict.items():
             result[BenchmarkType(framework)] = {}
-            for version, section in versions.items():
-                result[BenchmarkType(framework)][str(version)] = str(section)
+            for version, sections in versions.items():
+                if not isinstance(sections, list):
+                    raise Exception(f'section of benchmark {framework} version {version} is not of type list')
```
alonashkenazi, line 128: Should we move this change to a _verify_ function like the other checks?

Viperain: @alonashkenazi  i placed it there because i also do a parsing (from float to string) and I didn't want to run twice on the rules benchmarks.

alonashkenazi (approved):

PR #211, opened 2021-09-12
Feature/cr 3114 update origin func

idan-aviv:


cloudrail/knowledge/context/mergeable.py
```diff
@@ -13,6 +14,12 @@ class EntityOrigin(str, Enum):
     PSEUDO = 'pseudo'
 
 
+class UnSupportedEntityOriginType(Exception):
```
Tomerva, line 17: rename to UnSupportedEntityOriginTypeException

cloudrail/knowledge/context/mergeable.py
```diff
@@ -13,6 +14,12 @@ class EntityOrigin(str, Enum):
...
+
+    def __init__(self) -> None:
+        super().__init__('origin entity type don\'t supported')
```
Tomerva, line 20: change to 'IaC type {iac_type} not supported' and pass the iac_type = self.iac_state.iac_type

Tomerva (changes requested):

cloudrail/knowledge/context/mergeable.py
idan-aviv: @Tomerva the logic cause to exception is on origin entity not iac type

Tomerva: @idan-aviv but the error is originated from the value of self.iac_state.iac_type

cloudrail/knowledge/context/mergeable.py
```diff
@@ -5,12 +5,20 @@
...
 
 
 class EntityOrigin(str, Enum):
     LIVE_ENV = 'live_environment'
     TERRAFORM = 'terraform'
     PSEUDO = 'pseudo'
+    CLOUDFORMATION = 'cloudformation'
+
+
+class UnSupportedIacType(Exception):
```
Tomerva, line 18: rename to UnSupportedIacTypeException (all our exceptions end with Exception)

cloudrail/knowledge/context/mergeable.py
```diff
@@ -5,12 +5,20 @@
...
+
+    def __init__(self) -> None:
+        super().__init__('IaC type don\'t supported')
```
Tomerva, line 21: IaC type not supported

Tomerva (changes requested):

Tomerva (approved):

Tomerva (approved):

PR #219, opened 2021-09-19
CR-3043: Drift detection: Changing default SG for an EC2 instance is not detected

Imanuel-Mizrahi:
modified resource and class for EC2 instance

### DIFF
--- a/cloudrail/knowledge/context/aws/pseudo_builder.py
+++ b/cloudrail/knowledge/context/aws/pseudo_builder.py
@@ -144,7 +144,7 @@ class PseudoBuilder:
                 f'Could not find a subnet in the default vpc {default_vpc.vpc_id}'
             )
         else:
-            subnet = ResourceInvalidator.get_by_id(subnets, ec2.raw_data.subnet_id, True, ec2)
+            subnet = ResourceInvalidator.get_by_id(subnets, ec2.raw_data.subnet_id, False)
 
         eni_id = create_pseudo_id('eni')
         private_ip_address = ec2.raw_data.private_ip_address or subnet.cidr_block.split('/')[0]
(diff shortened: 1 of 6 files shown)


cloudrail/knowledge/context/aws/pseudo_builder.py
```diff
@@ -144,7 +144,7 @@ def create_ec2_network_interface(self, ec2: Ec2Instance, subnets: AliasesDict[Su
                 f'Could not find a subnet in the default vpc {default_vpc.vpc_id}'
             )
         else:
-            subnet = ResourceInvalidator.get_by_id(subnets, ec2.raw_data.subnet_id, True, ec2)
+            subnet = ResourceInvalidator.get_by_id(subnets, ec2.raw_data.subnet_id, False)
```
Tomerva, line 147: why?

cloudrail/knowledge/context/aws/resources/ec2/ec2_instance.py
```diff
@@ -70,7 +70,8 @@ def __init__(self,
                  tags: dict,
                  instance_type: str,
                  ebs_optimized: bool,
-                 monitoring_enabled: bool):
+                 monitoring_enabled: bool,
+                 security_groups_ids: List[str]):
```
Tomerva, line 74: the security groups are attached to the underlying network interface, not the actual EC2

Tomerva (changes requested):

cloudrail/knowledge/context/aws/resources/ec2/ec2_instance.py
Imanuel-Mizrahi: @Tomerva , but we agreed with @yi2020 that we should add the "security_group_ids" for the EC2 instance, as we had no other solution for drift.

cloudrail/knowledge/context/aws/pseudo_builder.py
Imanuel-Mizrahi: @Tomerva 
When you create a simple EC2, using TF, without any networking, AWS create all the networking by it's own.
This means that the TF context will not hold the subnet of this EC2, in it's TF context.
As a result, this function will invalidate the EC2, while it is not correct.
It will be correct to invalidate it only after the merged context, but not while building the drift context.

Imanuel-Mizrahi:
Will have a different solution once drift detection explicit will be ready, moving ticket to backlog

PR #221, opened 2021-09-29
cloudformation_iam_policies_builder, cloudformation_iam_role_builder:…

idan-aviv:
… new iam builders


cloudrail/knowledge/context/aws/resources/iam/role.py
```diff
@@ -37,12 +37,13 @@ def __init__(self, account: str,
...
         self.creation_date: str = creation_date
         self.last_used_date: RoleLastUsed = None
+        self.with_aliases(role_name, role_id, self.qualified_arn)
 
     def get_keys(self) -> List[str]:
         return [self.arn]
 
     def get_id(self) -> str:
-        return self.role_id
+        return self.role_name
```
Imanuel-Mizrahi, line 46: I do not really like the idea that the get_id returns the role_name and not the ID.

Imanuel-Mizrahi (commented):
Besides the usage of role_name into the ID, I think the rest is fine.

cloudrail/knowledge/context/aws/resources/iam/role.py
idan-aviv: @Imanuel-Mizrahi but role name is unique for the all account, besides that where in the code you actually use role id? even API calls using role name i.e. "aws iam get-role --role-name <value>"
why do you see it as a problem?

Imanuel-Mizrahi: It's not a functional problem, more that we use a function to return something else.
If you must use it, due to cloudformation - please do.

Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/aws/resources/iam/role.py
Tomerva: why did you need to change it in the fiirst place?

Tomerva (approved):

PR #280, opened 2021-10-27
enrich scanner context for cfn context builder

idan-aviv:


cloudrail/knowledge/context/aws/cloudformation/aws_cloudformation_context_builder.py
```diff
@@ -116,6 +119,9 @@ def build(iac_file: str,
             raise Exception('missing \'region\' parameter')
 
         if scanner_environment_context:
+            aws_assigner: AwsRelationsAssigner = AwsRelationsAssigner(scanner_environment_context)
+            aws_assigner.add_default_vpc_tasks_to_pool()
+            aws_assigner.run()
```
Viperain, line 124: @idan-aviv  

1. why not running all aws_assigner methods? 
how u decided what methods to enter? i prefer to run all aws_assigner methods.

2. if there is an issue with running the non-default methods i prefer to split the aws_assigner  to two classes (aws_assigner and aws_defaults_assigner) the enrichemnt will run both but cfn will run only aws_defaults_assigner. 

cc @Tomerva

Viperain (changes requested):

cloudrail/knowledge/context/aws/cloudformation/aws_cloudformation_context_builder.py
idan-aviv: @Viperain @Tomerva before i'm heading to fix some tests please take a look on the new class: AwsDefaultsRelationsAssigner

Viperain: @idan-aviv  as i wrote- if there is no issue with running the full assigned lets run the full assigner. no need to add extra choices to the developer where to add the assginer methods and assgienr is normally fast (the issue is with the enrichment port which will not happen)

cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -154,30 +154,28 @@
...
 from cloudrail.knowledge.context.environment_context.business_logic.resource_invalidator import ResourceInvalidator
+from cloudrail.knowledge.context.ip_protocol import IpProtocol
+from cloudrail.knowledge.context.mergeable import EntityOrigin, Mergeable
+from cloudrail.knowledge.utils.arn_utils import get_arn_resource
 from cloudrail.knowledge.utils.policy_evaluator import is_action_subset_allowed, PolicyEvaluator
 from cloudrail.knowledge.utils.policy_utils import is_policy_block_public_access
+from cloudrail.knowledge.utils.utils import flat_list, hash_list
+
 
+class AwsRelationsAssigner(AwsDefaultsRelationsAssigner):
```
Tomerva, line 173: dont think its right to use inheritance here, the relations assigner can **use** the default assigner, but its not a logical sub-part of it

cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -2675,3 +2562,18 @@ def get_attribute():
             return attribute
 
         global_accelerator.attributes = ResourceInvalidator.get_by_logic(get_attribute, False)
+
+    def add_default_vpc_tasks_to_pool(self):
```
Tomerva, line 2566: shouldnt this be removed?

cloudrail/knowledge/context/environment_context/business_logic/dependency_invocation.py
```diff
@@ -31,14 +31,19 @@ class CircularDependencyException(Exception):
...
 
 class DependencyInvocation:
-    def __init__(self, *function_pools: List[FunctionData], context: Optional[BaseEnvironmentContext] = None):
-        for function_pool in function_pools:
-            self._assert_no_circular_dependency(function_pool)
-            function_pool.sort(key=lambda x: len(x.dependencies))
 
-        self.function_pools: Tuple[List[FunctionData]] = function_pools
+    def __init__(self, context: Optional[BaseEnvironmentContext] = None):
+        self._function_pools: Set[List[FunctionData]] = set()
```
Tomerva, line 36: for performance it should be list/tuple instead of set

Tomerva (changes requested):

PR #298, opened 2021-11-10
compute instance no public ip rule

doronafota:


cloudrail/knowledge/rules/gcp/non_context_aware/compute_instance_no_public_ip_rule.py
```diff
@@ -0,0 +1,27 @@
...
+    def execute(self, env_context: GcpEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for compute_instance in env_context.compute_instances:
+            if network_interfaces_list := compute_instance.network_interfaces:
+                for network_interface in network_interfaces_list:
+                    if network_interface.access_config:
+                        issues.append(
+                            Issue(
+                                f"The {compute_instance.get_type()} `{compute_instance.get_friendly_name()}` has access_config sub block present",
+                                compute_instance,
```
Imanuel-Mizrahi, line 22: I had something similar in my rule lately.
I am not sure we need the evidence string to alert about the TF structure, but on the actual meaning.
Please advise with satya how to make the evidence more generic. 
Something like: 
"The {compute_instance.get_type()} `{compute_instance.get_friendly_name()}` has interfaces with public ip" 
@idan-aviv @Tomerva , what do you think?

Imanuel-Mizrahi (changes requested):

cloudrail/knowledge/rules/gcp/non_context_aware/compute_instance_no_public_ip_rule.py
Tomerva: yes, in general evidence should not talk about blocks. thats the remediation step's job

Tomerva (approved):

cloudrail/knowledge/rules/gcp/non_context_aware/compute_instance_no_public_ip_rule.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class ComputeInstanceNoPublicIpRule(GcpBaseRule):
+    def get_id(self) -> str:
+        return 'non_car_compute_instance_ensure_no_public_ip'
+
+    def execute(self, env_context: GcpEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for compute_instance in env_context.compute_instances:
+            if network_interfaces_list := compute_instance.network_interfaces:
```
idan-aviv, line 16: @doronafota redundant if condition

tests/knowledge/rules/gcp/non_context_aware/test_compute_instance_no_public_ip_rule.py
```diff
@@ -0,0 +1,43 @@
...
+            ["One public ip", "1.1.1.1", None, 1, True],
+            ["Both public ip", "1.1.1.1", "2.2.2.2", 2, True],
+            ["No public ip", None, None, 0, False],
+        ]
+    )
+
+    def test_compute_instance_public_ip(self, unused_name: str, public_ip_1: str, public_ip_2: str, isusses: int, should_alert: bool):
+        # Arrange
+        compute_instance_1 = create_empty_entity(GcpComputeInstance)
+        compute_instance_2 = create_empty_entity(GcpComputeInstance)
```
idan-aviv, line 25: @doronafota can you please explain the meaning of this condition?

tests/knowledge/rules/gcp/non_context_aware/test_compute_instance_no_public_ip_rule.py
```diff
@@ -0,0 +1,43 @@
...
+            compute_instance_1.network_interfaces[0].access_config[0].nat_ip = public_ip_1
+        if public_ip_2:
+            compute_instance_2.network_interfaces = [create_empty_entity(GcpComputeInstanceNetworkInterface)]
+            compute_instance_2.network_interfaces[0].access_config = [create_empty_entity(GcpComputeInstanceNetIntfAccessCfg)]
+            compute_instance_2.network_interfaces[0].access_config[0].nat_ip = public_ip_2
+        context = GcpEnvironmentContext(compute_instances=[compute_instance_1, compute_instance_2])
+        # Act
+        result = self.rule.run(context, {})
+        # Assert
+        if should_alert:
```
idan-aviv, line 38: @doronafota can remove this condition and add 'total_issues' and 'rule_status' params

cloudrail/knowledge/rules/gcp/non_context_aware/compute_instance_no_public_ip_rule.py
doronafota: compute_instance.network_interfaces is optional attribute so if it will be None and i remove the if condition i can get exception.

Tomerva: then just do `if compute_instance.network_interfaces`
another way, which also reduce nesting, is removing this condition entirely and in the for loop:
`for network_interface in compute_instance.network_interfaces or []`

by the way, how come network_interfaces is optional? can there be a compute instance (a VM?) that has no network interface? i find that odd..

cloudrail/knowledge/rules/gcp/non_context_aware/compute_instance_no_public_ip_rule.py
Tomerva: i think the evidence should be ' ... **h**as public ip address attached

Tomerva (changes requested):

cloudrail/knowledge/rules/gcp/non_context_aware/compute_instance_no_public_ip_rule.py
Imanuel-Mizrahi: @doronafota , seems like Tomer is right, according to [TF docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#network_interface), network_interface is a required field.
The context ticket got me confused while creating the resource, I will fix it in another PR, not this one.

doronafota: @Imanuel-Mizrahi do you built this resource? do you know if the network_interfaces is actually optional as written ?

Imanuel-Mizrahi: @doronafota , yes, I built it, and as mentioned in my previous comment, Tomer is right.
Do not change anything in the resource itself in this PR, I will do it in another one.

idan-aviv (approved):

Imanuel-Mizrahi (approved):

Tomerva (approved):

PR #299, opened 2021-11-11
CR-3558: dont crash when  qualified_arn is empty

Viperain:


cloudrail/knowledge/context/aws/resources/lambda_/lambda_function.py
```diff
@@ -94,6 +94,8 @@ def is_tagable(self) -> bool:
         return True
 
     def _get_simplified_arn(self) -> str:
+        if not self.qualified_arn:
```
Imanuel-Mizrahi, line 97: @Viperain , I think this might be a problem to return empty string for something which should be the "get_keys".
In which scenario you won't have the "qualified_arn"?

Tomerva: yeah this is a problem as the return value here is important for this resource, but should probably be handled in a different ticket

Tomerva (approved):

cloudrail/knowledge/context/aws/resources/lambda_/lambda_function.py
Viperain: correct, we should investiage when its happen again but I don't want to crash..

Imanuel-Mizrahi (approved):

PR #328, opened 2021-11-22
gcp compute ssl policy resource

doronafota:


cloudrail/knowledge/context/gcp/resources/compute/gcp_compute_ssl_policy.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @property
+    def is_tagable(self) -> bool:
+        return False
+
+    @property
+    def is_labeled(self) -> bool:
+        return False
+
+    def get_id(self) -> str:
```
Imanuel-Mizrahi, line 40: I think we should have get_name instead of get_id

cloudrail/knowledge/context/gcp/resources_builders/terraform/compute_ssl_policy_builder.py
```diff
@@ -0,0 +1,16 @@
...
+from cloudrail.knowledge.context.gcp.resources_builders.terraform.base_gcp_terraform_builder import BaseGcpTerraformBuilder
+
+
+class ComputeSslPolicyBuilder(BaseGcpTerraformBuilder):
+
+    def do_build(self, attributes: dict) -> GcpComputeSslPolicy:
+        name = attributes["name"]
+        min_tls_version = self._get_known_value(attributes, "min_tls_version", "TLS_1_0")
+        profile = self._get_known_value(attributes, "profile", "COMPATIBLE")
+        custom_features = self._get_known_value(attributes, "custom_features") if profile == 'CUSTOM' else None
```
Imanuel-Mizrahi, line 12: Why we have a condition for "custom_features" only inside the TF builder, and not in the scanner?

doronafota: i can remove this condition ...
terraform check it

Imanuel-Mizrahi: yes, please remove it, I think we should be alligned with same as being done in the scanner.

Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/gcp/resources/compute/gcp_compute_ssl_policy.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def get_name(self) -> Optional[str]:
+        return self.name
+
+    def get_cloud_resource_url(self) -> Optional[str]:
+        return f'{self._BASE_URL}/net-security/sslpolicies/details/{self.name}?project={self.project_id}'
+
+    def get_type(self, is_plural: bool = False) -> str:
+        if not is_plural:
+            return 'Compute SSL Policy Details'
```
Tomerva, line 48: redundant code - just return the string as its the same

Tomerva (approved):

PR #333, opened 2021-11-23
CR-3434: EC2 machines with separate network interfaces don't trigger SSH rules

Imanuel-Mizrahi:
added builders and rysder test


cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -643,11 +643,22 @@ def _assign_default_network_acl_rules_for_tf(nacl: NetworkAcl, vpcs: AliasesDict
...
 
-    @staticmethod
-    def _assign_network_interface_subnets(network_interface: NetworkInterface, subnets: AliasesDict[Subnet]):
+    def _assign_network_interface_subnets(self, network_interface: NetworkInterface, subnets: AliasesDict[Subnet]):
         network_interface.subnet = ResourceInvalidator.get_by_id(subnets, network_interface.subnet_id, True, network_interface)
         if not network_interface.primary_ip_address:
             network_interface.primary_ip_address = network_interface.subnet.cidr_block
+        if not network_interface.vpc_id:
+            network_interface.vpc_id = network_interface.subnet.vpc_id
+            network_interface.vpc = network_interface.subnet.vpc
```
Tomerva, line 652: you need to make sure that the function that assigns vpc to a subnet was called before this function (as dependency in IterFuncData)

cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -643,11 +643,22 @@ def _assign_default_network_acl_rules_for_tf(nacl: NetworkAcl, vpcs: AliasesDict
...
+        should_associate_public_ip = network_interface.subnet.map_public_ip_on_launch
```
Tomerva, line 653: 1. the variable name is misleading
2. its unnecessary
3. we should have similar logic elsewhere in this file for _should_associate_public_ip.. please check and use same function

tests/knowledge/scenarios/ryder_test/existing_resources/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: apply
```
Tomerva, line 1: ryder_test should be a normal scenario for network_interface context object.
also you should try to see if you can simplify the test by merging all the resources into one main.tf

Tomerva (changes requested):

cloudrail/knowledge/context/aws/aws_relations_assigner.py
Imanuel-Mizrahi: Added

cloudrail/knowledge/context/aws/resources/ec2/network_interface.py
```diff
@@ -120,7 +120,6 @@ def to_drift_detection_object(self) -> dict:
                 'subnet_id': self.subnet_id,
                 'primary_ip_address': self.primary_ip_address,
                 'secondary_ip_addresses': self.secondary_ip_addresses,
-                'public_ip_address': self.public_ip_address,
```
Tomerva, line 123: good catch.. but now that i think of it i think the other ip addresses should be left out of drift detection because i think they are dynamic unless they come from ElasticIP, in which case thats up to the EIP resource to determine its DD object

Tomerva (changes requested):

cloudrail/knowledge/context/aws/resources/ec2/network_interface.py
Imanuel-Mizrahi: Removed the other ip attributes as well

Tomerva (approved):

PR #342, opened 2021-12-03
CR-3643/CR-3644/CR-3645_AWS_rules_mapping_to_ISO/SOC2/CCM

oqf:
AWS rules have been mapped to ISO 27001:2017, SOC 2, and CCM 4.0


Viperain:
@oqf  notice that the pr failed due to unknown benchmarks. we have a static list of benchmarks so we can validate that the rule metadata is filled correctly. i added the new values to the list and now the tests passed. please notice for next time

Viperain (approved):

PR #353, opened 2021-12-08
CR-3493 + CR-2782: [GCP] [Context] Storage Bucket IAM + [CAR][GCP] Ensure that Cloud Storage bucket is not anonymously or publicly accessible

Imanuel-Mizrahi:
added builders and tests


cloudrail/knowledge/context/gcp/gcp_iam_actions.py
```diff
@@ -0,0 +1,35 @@
...
+        return updated_iam_policies
+
+    @staticmethod
+    def merge_bindings(bindings: List[GcpIamPolicyBinding]) -> List[Optional[GcpIamPolicyBinding]]:
+        merged_bindings = [binding for binding in bindings if binding.condition]
+        roles = {binding.role for binding in bindings if not binding.condition}
+        role_members_map = {}
+        for role in roles:
+            role_members = list({', '.join(flat_list([binding.members for binding in bindings if role == binding.role]))})
+            role_members_map.update({role: role_members})
```
Tomerva, line 32: why update and not role_members_map[role] = role_members:

cloudrail/knowledge/context/gcp/gcp_iam_actions.py
```diff
@@ -0,0 +1,35 @@
+from typing import List, Dict, Optional
+from cloudrail.knowledge.context.gcp.resources.iam.iam_access_policy import IamAccessPolicy, GcpIamPolicyBinding
+from cloudrail.knowledge.utils.utils import flat_list
+
+class IamActions:
+
+    @classmethod
+    def merge_iam_policies(cls, iam_policies: List[IamAccessPolicy]):
```
Tomerva, line 8: add return type

cloudrail/knowledge/context/gcp/resources_builders/terraform/iam_policy_builder.py
```diff
@@ -0,0 +1,91 @@
...
+        if condition_data := binding_dict.get('condition'):
+            condition = GcpIamPolicyCondition(expression=condition_data[0]['expression'],
+                                            title=condition_data[0]['title'],
+                                            description=get_known_value_func(condition_data[0], 'description'))
+        return condition
+
+    @staticmethod
+    def validate_iam_policies_config(iam_policies: List[IamAccessPolicy]):
+        policies_types = {policy.policy_type for policy in iam_policies}
+        if len(policies_types) > 1 and GcpIamPolicyType.AUTHORITATIVE in policies_types:
```
Tomerva, line 45: performance wise it would be better if you would remove line 44 and then in line 45 do this:
`if any(policy.policy_type == GcpIamPolicyType.AUTHORITATIVE  for policy in iam_policies)`

Tomerva (changes requested):

cloudrail/knowledge/context/gcp/resources_builders/terraform/iam_policy_builder.py
Imanuel-Mizrahi: But this condition is not true.
What I wanted to check is:
1. We have more than **one** type of policy.
2. One of the types is "AUTHORITATIVE"
Using your suggested code, I will crash even if I will have only AUTHORITATIVE type.

cloudrail/knowledge/context/gcp/gcp_iam_actions.py
```diff
@@ -0,0 +1,35 @@
...
+        updated_iam_policies: List[IamAccessPolicy] = []
+        policies_by_name: Dict[str, List[GcpIamPolicyBinding]] = {}
+        for policy in iam_policies:
+            if policy.resource_name not in policies_by_name:
+                policies_by_name[policy.resource_name] = policy.bindings
+            else:
+                policies_by_name[policy.resource_name].extend(policy.bindings)
+        for value in policies_by_name.values():
+            value = cls.merge_bindings(value)
```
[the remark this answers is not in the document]
Imanuel-Mizrahi: Once I do that, I encounter issues with the drift mechanism.
The list of bindings has different order, so the comparison of the object got screwed = it reports about drifts, which are not actual drifts, just the order is different.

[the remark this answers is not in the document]
Imanuel-Mizrahi: Found the issue, fixed

Tomerva (approved):

PR #354, opened 2021-12-09
CR-3682: [Azure] [Context] Data Lake Analytics Account

Imanuel-Mizrahi:
added data lake analytics account for azure resource


cloudrail/knowledge/context/azure/resources_builders/scanner/data_lake_store_builder.py
```diff
@@ -16,10 +16,13 @@ def do_build(self, attributes: dict) -> AzureDataLakeStore:
         encryption_type = encryption_type or ('ServiceManaged' if encryption_state == FieldActive.ENABLED else '')
         firewall_allow_azure_ips: FieldActive = FieldActive(properties.get('firewallAllowAzureIps'))
         firewall_state: FieldActive = FieldActive(properties.get('firewallState'))
```
idan-aviv, line 18: @Imanuel-Mizrahi why did you require to change it?

cloudrail/knowledge/context/azure/resources_builders/terraform/data_lake_store_builder.py
```diff
@@ -12,11 +12,14 @@ def do_build(self, attributes: dict):
         encryption_type: str = self._get_known_value(attributes, 'encryption_type') or ('ServiceManaged' if encryption_state == FieldActive.ENABLED else '')
         firewall_allow_azure_ips: FieldActive = FieldActive(attributes.get('firewall_allow_azure_ips', 'Enabled'))
         firewall_state: FieldActive = FieldActive(attributes.get('firewall_state', 'Enabled'))
```
idan-aviv, line 14: @Imanuel-Mizrahi why did you require to change it?
identity field is required and always with one 'SystemAssigned' possible value

cloudrail/knowledge/context/azure/resources/storage/azure_data_lake_analytics_account.py
```diff
@@ -0,0 +1,63 @@
...
+        return [self.get_id()]
+
+    def get_name(self) -> str:
+        return self.name
+
+    def get_type(self, is_plural: bool = False) -> str:
+        return 'Data Lake Analytics Account' + ('s' if is_plural else '')
+
+    def to_drift_detection_object(self) -> dict:
+        return {
```
idan-aviv, line 56: tags updated can lead to some security issue?

cloudrail/knowledge/context/azure/resources_builders/scanner/data_lake_store_builder.py
Imanuel-Mizrahi: @idan-aviv 
In my scenario I had data_lake_store resource, but there was no "identity" field at all!
Not in the TF template, and not in the scanner API call output, so I had to change that.

cloudrail/knowledge/context/azure/resources_builders/terraform/data_lake_store_builder.py
Imanuel-Mizrahi: @idan-aviv 
In my scenario I had data_lake_store resource, but there was no "identity" field at all!
Since the TF docs says that currently the only supported value is 'SystemAssigned' I put this as the default value to return if there is "identity" field, however, we also have the ability to have data in the future, if anything will be changed.

cloudrail/knowledge/context/azure/resources/storage/azure_data_lake_analytics_account.py
Imanuel-Mizrahi: @idan-aviv , it's not matter of security only, but some configuration changes.
If this field is not something the user can change, we do not need to check for drifts, otherwise - we should.

cloudrail/knowledge/context/azure/resources_builders/scanner/data_lake_store_builder.py
idan-aviv: @Imanuel-Mizrahi are you sure it passes apply command? identity is require field
see [here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/data_lake_store#identity)

cloudrail/knowledge/context/azure/resources/storage/azure_data_lake_analytics_account.py
idan-aviv: @Imanuel-Mizrahi configuration changes that can lead to security issues...this isn't the all idea?

idan-aviv (approved):

PR #362, opened 2021-12-19
CR-3686: [Azure] [Context] Logic App Workflow + CR-2389: [CAR] [Azure] Ensure Logic App Workflow has diagnostic logs enabled

Imanuel-Mizrahi:


cloudrail/knowledge/context/azure/resources_builders/terraform/logic_app_workflow_builder.py
```diff
@@ -0,0 +1,54 @@
...
+                                                                           'https://schema.management.azure.com/providers/Microsoft.Logic/schemas/2016-06-01/workflowdefinition.json#'),
+                                     workflow_version=self._get_known_value(attributes, 'workflow_version', '1.0.0.0'),
+                                     parameters=parameters,
+                                     workflow_parameters=workflow_parameters)
+
+    def get_service_name(self) -> AzureResourceType:
+        return AzureResourceType.AZURERM_LOGIC_APP_WORKFLOW
+
+    @staticmethod
+    def _get_access_control_rule(data: dict, key: str, known_value_func: Callable) -> Optional[LogicAppWorkflowAccessControlRules]:
```
Tomerva, line 51: why you nee to pass known_value_func and not just make this function non-static and then use the _get_known_value function directly? you can achieve that also by using @classmethod as _get_known_value is actually static

Tomerva (changes requested):

Tomerva (approved):

PR #370, opened 2021-12-22
CR-3672: [Azure] [Context] Virtual Machine Scale Set Extension + CR-2333: [CAR] [Azure] Ensure Virtual Machine Scale Set has diagnostic logs enabled

Imanuel-Mizrahi:
added builders and test for the resource


Imanuel-Mizrahi:
While working on this PR, worked also on CR-3673, CR-3674, CR-3677

cloudrail/knowledge/context/azure/azure_relations_assigner.py
```diff
@@ -171,3 +176,17 @@ def _assign_monitor_activity_log_alert_to_subscription(subscription: AzureSubscr
         subscription.monitor_activity_alert_log_list = ResourceInvalidator.get_by_logic(
             lambda: [monitor for monitor in monitor_activity_log_alert if subscription.get_id() in monitor.scopes],
             False)
+
+
+    @staticmethod
+    def _assign_extension_to_vmss(vmss: AzureVirtualMachineScaleSet, vms_extentions: AliasesDict[AzureVirtualMachineExtension]):
```
Tomerva, line 182: you can improve this functions performance by running for each extension instead of for each vmss

then instead of using get_by_logic you can use get_by_id

cloudrail/knowledge/context/azure/azure_relations_assigner.py
```diff
@@ -171,3 +176,17 @@ def _assign_monitor_activity_log_alert_to_subscription(subscription: AzureSubscr
...
+        vmss.extensions = ResourceInvalidator.get_by_logic(
+            lambda: [extension for extension in vms_extentions if vmss.get_id().lower() == extension.attached_resource_id.lower()],
+            False)
+
+
+    @staticmethod
+    def _assign_extension_to_vm(vm: AzureVirtualMachine, vms_extentions: AliasesDict[AzureVirtualMachineExtension]):
```
Tomerva, line 189: same as above

tests/knowledge/rules/azure/context_aware/test_vmss_diagnostics_logs_enabled_rule.py
```diff
@@ -0,0 +1,44 @@
...
+    def setUp(self):
+        self.rule = VmssDiagnosticsLogsEnabledRule()
+
+    @parameterized.expand(
+        [
+            ["no_logs_enabled", False, True],
+            ["logs_enabled", True, False],
+        ]
+    )
+    def test_vmss_diagnostics_logs(self, unused_name: str, extnesion_configured: bool, should_alert: bool):
```
Tomerva, line 24: add tests for the different OS types, publishers, and extension types

Tomerva (changes requested):

tests/knowledge/rules/azure/context_aware/test_vmss_diagnostics_logs_enabled_rule.py
Imanuel-Mizrahi: Basically, as far as I understand, we should more invest the effort on the TF tests using the main.tf files, and there we have enough scenarios.
If you think I should also make some scenarios here, will do.

Tomerva: its not alot of effort to add the test because it just means adding more parameters above and making it generic

Tomerva (approved):

PR #372, opened 2021-12-23
non_car_storage_account_network_access_virtual_network_rules

doronafota:


cloudrail/knowledge/context/azure/resources/storage/azure_storage_account_network_rules.py
```diff
@@ -22,20 +24,32 @@ class BypassTrafficType(str, Enum):
     AZURESERVICES = 'AzureServices'
 
 
+@dataclass
+class AzureVirtualNetworkSubnetId:
```
Tomerva, line 28: this is an overkill.. why do we need a dataclass for using an id? is this how the ticket describes it?

cloudrail/knowledge/rules/azure/non_context_aware/ensure_storage_account_allows_network_access_using_vnet_rule.py
```diff
@@ -0,0 +1,25 @@
...
+
+class EnsureStorageAccountAllowsNetworkAccessUsingVnetRule(AzureBaseRule):
+
+    def get_id(self) -> str:
+        return 'non_car_storage_account_network_access_virtual_network_rules'
+
+    def execute(self, env_context: AzureEnvironmentContext, parameters: Dict[ParameterType, any]) -> List[Issue]:
+        issues: List[Issue] = []
+        for storage_account in env_context.storage_accounts:
+            if storage_account.network_rules.ip_rules or not storage_account.network_rules.virtual_network_subnet_ids:
```
Imanuel-Mizrahi, line 16: Something a bit strange to me between the evidence string and the rule logic:
We consider a violation if we do not have ip_rules and not virtual_network_subnets, but the evidence says that we are allowing network access based on IP rules.
So, how can we say there is no IP rules, but alert that we allow access using it?

Tomerva, line 16: are you sure about this condition?

doronafota: I asked Oscar and he said ip_rules and virtual_network_subnet_ids cant be applied together so i will change the condition to:  
if not storage_account.network_rules.virtual_network_subnet_ids:

doronafota: look my answer to Tomer

Imanuel-Mizrahi (approved):

cloudrail/knowledge/context/azure/resources/storage/azure_storage_account_network_rules.py
doronafota: no, i added this attribute to existing resource. 
'virtual network subnet id' object have more attributes that i didn't implement in this ticket because it's not necessary for me. maybe in the future someone will implement the other attributes so i thinking it will be better using here dataclass. 
what do you think?

Tomerva: what other attributes? it sounds like if there are other attributes then the name of the class is not appropriate

doronafota: the other attributes are: allow and state (related to the network rule of the specific subnet)

cloudrail/knowledge/rules/azure/non_context_aware/ensure_storage_account_allows_network_access_using_vnet_rule.py
Tomerva: with taking the evidence into consideration, it sounds like the condition should be
`if storage_account.network_rules.ip_rules`

Tomerva (approved):

PR #377, opened 2021-12-28
CR-3699: [Azure] [Context] MSSQL Server (and some other related SQL resources)

Imanuel-Mizrahi:
azure context mssql server


cloudrail/knowledge/context/azure/resources/databases/azure_sql_server.py
```diff
@@ -35,8 +96,18 @@ def get_friendly_name(self) -> str:
 
     @property
     def is_tagable(self) -> bool:
-        return False
+        return True
+
+    def get_type(self, is_plural: bool = False) -> str:
+        return 'Asure SQL server ' + ('s' if is_plural else '')
+
+    def get_managed_identities(self) -> List[AzureManagedIdentity]:
```
Tomerva, line 104: i dont see the point in get_managed_identities / get_managed_identities_ids

usually you create a getter method for an attribute you want to protect from writing (mostly so only this class or class that inherit from it are able to access it), so you will have get_x, but without the ability to set x

here, you can get the resource directly, or via this getter method, and also set it directly. so these methods are pointless

Tomerva (changes requested):

Tomerva (approved):

PR #383, opened 2021-12-30
CR-2698: [AWS] [CFN] test_ec2

Imanuel-Mizrahi:


cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -675,6 +676,16 @@ def _assign_security_group_rules(security_group: SecurityGroup, security_group_r
                      and sgr.connection_type == ConnectionType.OUTBOUND],
             False))
 
+    ## Added following structure in CFN:
+    ## https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html#:~:text=CidrIp%3A%200.0.0.0%2F0-,remove%20the%20default%20rule,-When%20you%20specify
+    @staticmethod
+    def _assign_default_security_group_rule(security_groups: AliasesDict[SecurityGroup]):
+        for security_group in security_groups:
+            if security_group.origin == EntityOrigin.CLOUDFORMATION and not security_group.outbound_permissions:
```
idan-aviv, line 684: @Imanuel-Mizrahi how it is different from TF? it shouldn't be the same behavior?

cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -849,7 +860,7 @@ def get_assume_role_policies():
     def _assign_iam_roles_policies(role: Role, policies: List[ManagedPolicy],
                                    policy_role_attachments: List[PolicyRoleAttachment]):
         def get_role_policies():
-            policy_arns = [pra.policy_arn for pra in policy_role_attachments if pra.role_name == role.role_name]
```
idan-aviv, line 852: @Imanuel-Mizrahi why role name changed to aliases?

cloudrail/knowledge/context/aws/resources_builders/cloudformation/ec2/cloudformation_ec2_builder.py
```diff
@@ -19,7 +19,7 @@ def parse_resource(self, cfn_res_attr: dict) -> Ec2Instance:
         network_interfaces = self.get_property(properties, 'NetworkInterfaces', [])
         network_interface_ids = [ni.get('NetworkInterfaceId') for ni in network_interfaces]
         network_interface_ids = network_interface_ids if is_iterable_with_values(network_interface_ids) else self.CFN_PSEUDO_LIST
-        security_group_ids_from_enis = set(flat_list([ni['GroupSet'] for ni in network_interfaces if ni.get('GroupSet')]))
```
idan-aviv, line 22: @Imanuel-Mizrahi please explain this change convert to: list->set->list

tests/knowledge/context/base_context_test.py
```diff
@@ -93,7 +93,7 @@ def run_test_case(self, module_path: str, assert_func, base_scanner_data_for_iac
         module_path = '{}/{}'.format(self.get_component(), module_path)
         if test_options.run_drift_detection:
             self._run_drift_detection_for_terraform(module_path)
-            self._run_drift_detection_for_cloudformation(module_path, test_options.cfn_template_params)
```
idan-aviv, line 96: uncomment

cloudrail/knowledge/context/aws/cloudformation/aws_cloudformation_context_builder.py
```diff
@@ -222,6 +223,8 @@ def build(iac_file: str,
             kinesis_streams=CloudformationKinesisStreamBuilder(cfn_by_type_map).build(),
             origin_access_identity_list=CloudformationCloudfrontOriginAccessIdentityBuilder(cfn_by_type_map).build(),
             s3_bucket_acls=CloudformationS3BucketAclBuilder(cfn_by_type_map).build(),
```
idan-aviv, line 225: @Imanuel-Mizrahi shouldn't be new context test cases for those builders?

cloudrail/knowledge/context/aws/aws_relations_assigner.py
Imanuel-Mizrahi: I just tested it to make sure.
When you create a security group using TF, you can create it without ANY rules, and AWS will not fill it in for you.
But when you create one using CFN - AWS will automatically create this rule, as specified in the document I mentioned in the relations assigner, just before this function.
I actually found it while running CFN DD.

cloudrail/knowledge/context/aws/aws_relations_assigner.py
Imanuel-Mizrahi: Because when running with CFN, the role name was not the actual role name, it was the resource CFN name (physical name?), I found that it was part of the aliases, so I used it, to cover all cases.

cloudrail/knowledge/context/aws/resources_builders/cloudformation/ec2/cloudformation_ec2_builder.py
Imanuel-Mizrahi: using as a set - to remove duplicates.
Turning it back to a list - because otherwise there is a different type for this attribute - it failed the DD test for me.

tests/knowledge/context/base_context_test.py
Imanuel-Mizrahi: Correct, my mistake, will change it ASAP.

cloudrail/knowledge/context/aws/cloudformation/aws_cloudformation_context_builder.py
Imanuel-Mizrahi: Probably there should, but I just focused on EC2 tests, and I added the needed resources for my case.
There should be tickets for each one later.
Do you want me to take the other tickets, and work on it as part of this one?

cloudrail/knowledge/context/aws/aws_relations_assigner.py
idan-aviv: @Imanuel-Mizrahi the real question is default security group do contains default outbound rule? in the actual environment? (no matter if AWS return it via API or not)

Imanuel-Mizrahi: yes, I saw it in the console while applying CFN.
Didn't see in the console, while applying with TF
In short: we have different behavior between CFN and TF

Tomerva: please add a comment explaining this

cloudrail/knowledge/context/aws/aws_relations_assigner.py
Tomerva: but is the role_name an alias? is it unique enough?

idan-aviv: in case of RoleName don't set explicitly the logical id will be the role name, so why "pra.role_name == role.role_name" don't find any matching?

cloudrail/knowledge/context/aws/cloudformation/aws_cloudformation_context_builder.py
idan-aviv: i think it is the correct way if we are adding new builders and want to be sure it's covered

Tomerva: probably would be best, so if there are issues youll find them now

cloudrail/knowledge/context/aws/aws_relations_assigner.py
Imanuel-Mizrahi: @idan-aviv , cannot remember exactly, I know that is was not equal to the role_name, but only part of the aliases.
This is the CFN template:
```
  S3FullAccessRole:
    Type: AWS::IAM::Role
    Properties:
      RoleName: "s3-full-access-role"
      AssumeRolePolicyDocument:
        Version: '2012-10-17'
        Statement:
        - Effect: Allow
          Principal:
            AWS: "*"
          Action:
          - sts:AssumeRole

  Policy:
    Type: AWS::IAM::Policy
    Properties:
      PolicyName: "allow_role_s3_operations"
      PolicyDocument:
        Version: "2012-10-17"
        Statement:
          - Effect: Allow
            Action:
              - "s3:*"
            Resource: "*"
      Roles:
        - !Ref S3FullAccessRole
```
Inside the "policy" resource, you can see that the "Roles" is using Ref to the role resource.
What eventually I get into the role_name (inside the policy resource) is the value: "S3FullAccessRole", and not the role name ("s3-full-access-role").
I did find this value inside the aliases.
As the aliases, should be unique per resource, I didn't thought this should be an issue.

cloudrail/knowledge/context/aws/resources_builders/cloudformation/ec2/cloudformation_ec2_builder.py
idan-aviv: suggestion: you can add each sg id to set from the beginning and than convert it to list...your call

cloudrail/knowledge/context/aws/cloudformation/aws_cloudformation_context_builder.py
Imanuel-Mizrahi: OK, will work on that, will let you know once ready

Tomerva (changes requested):

cloudrail/knowledge/context/aws/aws_relations_assigner.py
idan-aviv: ohhh o.k this test case ran without cfn-account-data so it is reasonable

cloudrail/knowledge/context/aws/resources/iam/iam_group.py
```diff
@@ -16,11 +16,14 @@ def __init__(self, account: str, name: str, group_id: str, qualified_arn: str, a
         self.group_id: str = group_id
 
     def get_keys(self) -> List[str]:
-        return [self.group_id]
+        return [self.account, self.name]
```
Tomerva, line 19: why did you need to change this? group_id is a perfectly good key

cloudrail/knowledge/context/aws/resources/iam/iam_group.py
```diff
@@ -16,11 +16,14 @@ def __init__(self, account: str, name: str, group_id: str, qualified_arn: str, a
...
 
     def get_name(self) -> str:
         return self.name
 
+    def get_id(self) -> str:
+        return self.name
```
Tomerva, line 25: im not sure name is unique enough to be qualified as an id

cloudrail/knowledge/context/aws/resources/iam/iam_user.py
```diff
@@ -33,11 +33,14 @@ def get_policies_attach_origin_maps(self) -> List[Dict]:
         return self.policy_attach_origin_map + flat_list([group.policy_attach_origin_map for group in self.groups])
 
     def get_keys(self) -> List[str]:
-        return [self.user_id]
```
Tomerva, line 36: why did you need to change this?

cloudrail/knowledge/context/aws/resources/iam/iam_user.py
```diff
@@ -33,11 +33,14 @@ def get_policies_attach_origin_maps(self) -> List[Dict]:
...
+        return [self.account, self.name]
 
     def get_arn(self) -> str:
         return self.qualified_arn
 
+    def get_id(self) -> str:
```
Tomerva, line 41: im not sure name is unique enough to be qualified as an id

cloudrail/knowledge/context/aws/resources/iam/policy_user_attachment.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, account: str, policy_arn: str, user_id: str, user_name: str):
         self.user_name: str = user_name
 
     def get_keys(self) -> List[str]:
-        return [self.policy_arn, self.user_id]
+        return [self.policy_arn, self.user_name, self.account]
```
Tomerva, line 22: arn should be enough.. why change?

Tomerva (changes requested):

cloudrail/knowledge/context/aws/resources/iam/iam_group.py
Imanuel-Mizrahi: Answered internally, CFN does not provide us with the ID.

cloudrail/knowledge/context/aws/resources/iam/iam_group.py
Imanuel-Mizrahi: How much the value should be unique?
It's just that in CFN we need to return inside the "get_id" the value which CFN returns for the "Ref":
https://indeni.atlassian.net/wiki/spaces/RnD/pages/2330001409/Developing+new+CFN+builder+guide

cloudrail/knowledge/context/aws/resources/iam/iam_user.py
Imanuel-Mizrahi: Answered internally, CFN does not provide us with the ID.

cloudrail/knowledge/context/aws/resources/iam/iam_user.py
Imanuel-Mizrahi: How much the value should be unique?
It's just that in CFN we need to return inside the "get_id" the value which CFN returns for the "Ref":
https://indeni.atlassian.net/wiki/spaces/RnD/pages/2330001409/Developing+new+CFN+builder+guide

cloudrail/knowledge/context/aws/resources/iam/policy_user_attachment.py
Imanuel-Mizrahi: Answered internally, CFN does not provide us with the ID.

Tomerva (approved):

idan-aviv (approved):

PR #390, opened 2022-01-05
CR-3343: [AWS] [CFN] test_rds

Imanuel-Mizrahi:


cloudrail/knowledge/context/aws/scanner/aws_scanner_context_builder.py
```diff
@@ -197,7 +197,9 @@ def build(account_data_dir: Optional[str], account_id: Optional[str] = None, sal
                 main_route_table_associations=MainRouteTableAssociationBuilder(account_data_dir, salt).build(),
                 kms_aliases=KmsAliasBuilder(account_data_dir, salt).build(),
                 resources_tagging_list=ResourceTagMappingListBuilder(account_data_dir, salt).build(),
-                origin_access_identity_list=OriginAccessIdentityBuilder(account_data_dir, salt).build())
+                origin_access_identity_list=OriginAccessIdentityBuilder(account_data_dir, salt).build(),
+                internet_gateways=[igw for igw in InternetGatewayBuilder(account_data_dir, salt).build()
+                                   if any(igw.vpc_id == vpc.vpc_id for vpc in vpcs)])
```
Tomerva, line 202: why do you need this condition?

Tomerva (changes requested):

cloudrail/knowledge/context/aws/scanner/aws_scanner_context_builder.py
Imanuel-Mizrahi: I wanted to collect only the internet GW of the default VPC, not all IGW's.

Imanuel-Mizrahi:
Note about RDS instance engine_version value:
## Engine version is tricky, it depends on the engine_type, but for every type there is a different default version, and also depends on the region.
## There is an API call to get the default.
## Currently, I have removed "engine_version" from DD, we will fetch this data from the scanner context.
## Although "engine_version" attribute could be changed for RDS instance, but not from the RDS cluster.

cloudrail/knowledge/context/aws/scanner/aws_scanner_context_builder.py
Tomerva: but we need igw for other vpc's....

Tomerva (approved):

idan-aviv (approved):

PR #392, opened 2022-01-06
CR-3378: [AWS] [CFN] test_eks

Imanuel-Mizrahi:
modified cfn builders and test scenarios


Imanuel-Mizrahi:
I have removed a few fields from the dd object of EKS, since you can barely change any field after creation.

cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -1495,11 +1495,14 @@ def _assign_eks_cluster_eni(self,
                                 security_groups: AliasesDict[SecurityGroup]):
         self._assign_network_configuration_to_eni(eks_cluster, eks_cluster.get_all_network_configurations(), subnets, False)
 
-        if eks_cluster.is_managed_by_iac and eks_cluster.cluster_security_group_id is None:
+        if eks_cluster.is_managed_by_iac and (eks_cluster.cluster_security_group_id is None or not eks_cluster.security_group_ids):
             security_group = self.pseudo_builder.create_security_group(eks_cluster.network_resource.vpc, False,
                                                                        eks_cluster.account, eks_cluster.region)
             security_groups.update(security_group)
```
Tomerva, line 1501: create_security_group already updates the context

cloudrail/knowledge/context/aws/resources/eks/eks_cluster.py
```diff
@@ -49,7 +49,8 @@ def __init__(self,
         self.role_arn: str = role_arn
         self.arn: str = arn
         self.endpoint: str = endpoint
-        security_groups = security_group_ids
+        self.security_group_ids: List[str] = security_group_ids
+        security_groups = self.security_group_ids
```
Tomerva, line 53: this initialization is redundant, please remove it
if the variable name was accurate it would be kinda fine, but the name is misleading, and creating an assignment for it doesnt really reduce amount of code

Tomerva (changes requested):

cloudrail/knowledge/context/aws/resources/eks/eks_cluster.py
Imanuel-Mizrahi: Turns out that I need it, just because I added a logic in the relations assigner to check the value of this attribute.
Otherwise the code crashes...
Adding it back

cloudrail/knowledge/context/aws/aws_relations_assigner.py
```diff
@@ -1489,17 +1488,16 @@ def _assign_elastic_search_domain_subnets(self,
...
 
-    def _assign_eks_cluster_eni(self,
-                                eks_cluster: EksCluster,
-                                subnets: AliasesDict[Subnet],
-                                security_groups: AliasesDict[SecurityGroup]):
+    def _assign_eks_cluster_eni(self, eks_cluster: EksCluster, subnets: AliasesDict[Subnet]):
         self._assign_network_configuration_to_eni(eks_cluster, eks_cluster.get_all_network_configurations(), subnets, False)
 
-        if eks_cluster.is_managed_by_iac and eks_cluster.cluster_security_group_id is None:
+        if eks_cluster.is_managed_by_iac and (eks_cluster.cluster_security_group_id is None or not eks_cluster.security_group_ids):
```
Tomerva, line 1494: are you sure about this? this means that if he has a cluster SG but not "regular" SG's then it will create a pseudo one.. not sure this is what happens

Tomerva (changes requested):

cloudrail/knowledge/context/aws/aws_relations_assigner.py
Imanuel-Mizrahi: I think you are right.
Based on this link:
https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html#:~:text=in%20their%20description.-,Note,-If%20you%20used
Sounds like it will take the default VPC SG.
Will try to change the logic accordingly once I can.

Imanuel-Mizrahi: Done

Tomerva (approved):

idan-aviv (approved):

PR #396, opened 2022-01-09
CR-3767: [Azure] [Context] Network Interface NAT Rule Association

Imanuel-Mizrahi:
modified resource, builders, tests


cloudrail/knowledge/context/azure/resources/network/azure_network_interface_nat_rule_association.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def get_cloud_resource_url(self) -> Optional[str]:
+        pass
+
+    @property
+    def is_tagable(self) -> bool:
+        return False
+
+    def get_keys(self) -> List[str]:
+        return [self.network_interface_id, self.nat_rule_id]
```
Tomerva, line 31: maybe ip_configuration_name should also be included? sounds like you maybe you are able to set up the same association (same NIC and NAT rule) to different configs

Tomerva (approved):

